PR #53, opened 2014-07-18
Various small fixes and one big fix in the scripting SM event handling code

psoetens:


meyerj:
This patch and also the original version that defines `BOOST_DATE_TIME_POSIX_TIME_STD_CONFIG` in public headers is invalid.

`BOOST_DATE_TIME_POSIX_TIME_STD_CONFIG` is a compile-time option of Boost, which should not be defined in a header file. You might get away with defining it in a cpp file and only use Boost header-only functions. But other libraries like [Boost.Thread](http://www.boost.org/doc/libs/1_66_0/doc/html/thread.html) or string formatting in [Boost.DateTime](http://www.boost.org/doc/libs/1_66_0/doc/html/date_time.html) are typically pre-compiled and the differences of the internal representation of `boost::posix_time::ptime` caused by defining `BOOST_DATE_TIME_POSIX_TIME_STD_CONFIG` might cause run-time errors or unexpected behavior if the user mixes RTT code with calls into one of those Boost libraries.

See http://www.boost.org/doc/libs/1_66_0/more/getting_started/unix-variants.html#header-only-libraries for an overview of which Boost libraries are header-only.

PR #66, opened 2014-10-14
Corba aliases

smits:
This PR allows to create CORBA component servers that instead of using their own name, use an alias.  This is handy if you have X similar deployments in which you do not want to do component renaming for various reasons but you still want to remotely connect to multiple of them.  In this case registering them under an alias allows you to create remote proxies that remotely will be known under their alias name.


psoetens:
I agree that this needs refactoring...

smits:
I refactored and tested again. Please reconsider merging ;)

snrkiwi:
While I agree with _meyerj_ that _comps_ may not be needed, it seems that unwinding it from all the deployment code could be a bad thing. It's used in almost 80 different places. I'd argue for leaving it in for now, and perhaps dealing with it later?

smits:
I agree with @snrkiwi , refactoring out the comps could be part of another PR.

meyerj:
merged into toolchain-2.9: 5699152daf57c776c8f49b1a644df2686e48fd80

PR #70, opened 2014-10-22
updateHook() called too often

psoetens:
As discussed in length in this forum thread: http://www.orocos.org/forum/rtt/rtt-dev/changing-behaviour-updatehook :

We have observed lately two major issues:

> 1. updateHook() is called *too many times* (even according to
> specifications), especially when 'OwnThread' operations are implemented OR
> called(!).
> -> This was due to internal bookkeeping of the ExecutionEngine to dispatch
> the asynchronous requests. Users found it confusing that updateHook was
> called during the use of operations. This also kills performance of Lua
> state machines
> sitting in updateHook(), since they are evaluated too many times.
>
> 2. Installing a callback for an event port caused calling that callback AND
> updateHook(), instead of calling only the callback.
> -> The callback does not replace the call to updateHook(). This is also
> related to users being surprised that an event port callback is only called
> periodically in periodic threads, while they thought it would be called
> immediately upon each event reception, and eventually updateHook
> periodically (but serialized with the callbacks).
>
> The idea then rose to have updateHook() to only be called after:
> 1) any call to TaskContext::trigger()
> -> for activities which allow this user control
> 2) the component's activity triggered itself (periodic, file descriptor,
> ...)
> -> what users expect
> 3) the component is being update()
> -> for slave activity behavior

I'm creating a simple patch that addresses both issues, in a backwards compatible way.

### DIFF
--- a/rtt/DataFlowInterface.cpp
+++ b/rtt/DataFlowInterface.cpp
@@ -153,6 +153,8 @@ namespace RTT
 #endif
         if (callback)
             mservice->getOwner()->dataOnPortCallback(&port,callback); // the handle will be deleted when the port is removed.
+        else
+            mservice->getOwner()->dataOnPortCallback(&port,boost::bind(&TaskCore::trigger, mservice->getOwner()) ); // default schedules an updateHook()
 
 #ifndef ORO_SIGNALLING_PORTS
         port.signalInterface(true);
--- a/rtt/DataFlowInterface.hpp
+++ b/rtt/DataFlowInterface.hpp
@@ -107,9 +107,12 @@ namespace RTT
          * add a Service with the same name of the port.
          * @param name The name to give to the port.
          * @param port The port to add.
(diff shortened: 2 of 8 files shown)


ahoarau:
I'm seeing a very strange problem where my updatehook is called too often too.
I set a periodic component at HighestPriority + ORO_SCHED_RT, period of 2ms.
Inside this component I'm computing the period : 
```c++
    update_period_s = TimeService::Instance()->secondsSince(last_updatehook_ticks);
    last_updatehook_ticks = TimeService::Instance()->getTicks();
```
Period is perfect for a while (like 2ms +- 0.01ms for ex).
I set the system under load.
After a random amount of time the computed period becomes 0.2ms (!!), like if something was triggering the updatehook(), making the CPU goes to 100%. There is no `trigger()` inside this component.
I'm using the latest 2.9 branch.

ahoarau:
To add more info : the component is doing nothing but computing the period (so the update hook computation time is very close to zero).

PR #81, opened 2015-01-06
Allow the construction of an uninitialized TaskContextProxy

jmachowinski:
This is useful if a derived class wants to initialize the
typekits, before initializing the proxy itself.

The current usecase is the orogen cpp proxy layer, were
proxies for tasks are generated. In this case we want to
initialize the typekit for the task in the constructor of the 
orogen-proxy before synchronizing. 

This is the same merge request as for master
orocos-toolchain/rtt#80
but for the rock1408 branch.

### DIFF
--- a/rtt/transports/corba/TaskContextProxy.cpp
+++ b/rtt/transports/corba/TaskContextProxy.cpp
@@ -94,6 +94,16 @@ namespace RTT
     TaskContextProxy::TaskContextProxy(std::string name, bool is_ior)
         : TaskContext("NotFound")
     {
+        initFromURIOrTaskname(name, is_ior);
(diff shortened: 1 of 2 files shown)


jmachowinski:
As nobody replied this, I assume it is fine, I will merge it in 2 days

psoetens:
I don't understand the merge intent here. Do you want to merge into orocos-toolchain:master or into orocos-toolchain:rock1408 ?

jmachowinski:
This one should merge into rock1408

jmachowinski:
Ahh I set it up wrong. As the merge in general seems to be fine, I close this one and push it manually

PR #82, opened 2015-01-12
Add connection introspection ability

jmachowinski:
Hi, 
this patch set adds the ability to introspect the connections that are active.
This feature is mainly useful for debugging. For now it is based on the 
rock1408 branch and SHOULD NOT BE MERGED.

Feel free to give feedback, if everyone is fine with this, I will rebase the
code on the current master and recreate a pull request.

By the way, it seems as  CDataFlowInterface_i::buildChannelInput is dead code.
Greetings
   Janosch

### DIFF
--- a/rtt/base/ChannelElementBase.hpp
+++ b/rtt/base/ChannelElementBase.hpp
@@ -70,6 +70,11 @@ namespace RTT { namespace base {
         shared_ptr input;
         shared_ptr output;
 
+        std::string inputPortName;
+        std::string inputTaskName;
+        std::string outputPortName;
+        std::string outputTaskName;
+        
         RTT::os::Mutex inout_lock;
 
     protected:
@@ -88,6 +93,40 @@ namespace RTT { namespace base {
         ChannelElementBase();
         virtual ~ChannelElementBase();
 
+        void setInputPortName(const std::string &name)
+        {
+            inputPortName = name;
+        };
+        void setInputTaskName(const std::string &name)
+        {
+            inputTaskName = name;
+        };
+        void setOutputPortName(const std::string &name)
+        {
+            outputPortName = name;
+        };
+        void setOutputTaskName(const std::string &name)
+        {
+            outputTaskName = name;
+        };
+        
+        const std::string &getInputPortName() const
+        {
+            return inputPortName;
+        };
+        const std::string &getInputTaskName() const
+        {
+            return inputTaskName;
+        };
+        const std::string &getOutputPortName() const
+        {
+            return outputPortName;
+        };
+        const std::string &getOutputTaskName() const
+        {
+            return outputTaskName;
+        };
+        
         /**
          * Removes the input channel (if any).
          * This call may delete channels from memory.
--- a/rtt/base/ChannelElementBase.hpp
+++ b/rtt/base/ChannelElementBase.hpp
@@ -47,6 +47,7 @@
 #include <rtt/os/Mutex.hpp>
 #include "rtt-base-fwd.hpp"
 #include "../internal/rtt-internal-fwd.hpp"
+#include <string>
 
 namespace RTT { namespace base {
 
--- a/rtt/base/InputPortInterface.cpp
+++ b/rtt/base/InputPortInterface.cpp
@@ -109,6 +109,10 @@ bool InputPortInterface::channelReady(ChannelElementBase::shared_ptr channel, RT
     // cid is deleted/owned by the ConnectionManager.
     if ( channel ) {
         internal::ConnID* cid = channel->getConnID();
+        cid->setOutputPortName(channel->getOutputPortName());
+        cid->setOutputTaskName(channel->getOutputTaskName());
+        cid->setInputPortName(channel->getInputPortName());
+        cid->setInputTaskName(channel->getInputTaskName());
         if (cid ) {
             this->addConnection(cid, channel, policy);
             if ( channel->inputReady() )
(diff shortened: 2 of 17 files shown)


doudou:
Thanks for looking into that, that is a much needed feature.

The main drawback in your method is that it is really only useful for debugging. There is no one-to-one mapping between task names and the name that can be resolved by the tooling (obviously in Rock since we have multiple name services, but also in RTT since they are introducing using different names between the task name stored in TaskContext objects and the task name as registered on the name server)

I had in mind the idea to implement it. I thought that the best method would be to store this information in the connection policy, and allowing the callers to enumerate the policies of the channels currently connected to a port. Since the policy is already passed along during connection, the change in the internal API would be minimal. And this is already the way publish/subscribe architectures are handled (e.g. the ROS topic name is stored in the connection policy). 

What you would then need to do is auto-fill the fields if they are empty to make sure that *some* information is always there. It would cover your use case and the one where the tooling would properly set the fields so that tasks can be resolved by the tooling.

jmachowinski:
The basic idea here is the same. To get the connected tasks, you get the ConnManager, and enumerate over the ChannelDescriptors. Instead of storing the connection information inside the policy I went for the ConnId, as it looked like the more logical place.
 
You don't really win much by putting the information into the policy thou. In the case of the 
port->connect(otherPort) you would still need to gather the information about the tasks. Note that
the core RTT api change is minimal. The most changes needed to be done inside of corba.

For the name thing, sure you have to go for canonical task names but I don't see a big problem 
with this. If this is really an issue one might enhance the ConnId by a URI or something.

doudou:
Basically, the difference in perception between what you propose and what I propose is that RTT should by design be completely separated from the transport concerns. Putting the information in the policy is the most flexible choice under this constraint, as the framework (which concerns itself about name services and such) is the one putting the information in the URI, not RTT.

Whether ConnID is the most logical place is arguable. The policy is storing per-connection metadata, which is exactly what we're trying to achieve.

Moreover, putting it in the policy makes the changes a lot smaller as you don't need to add code to propagate the information (the policy is already propagated).

jmachowinski:
The introspection patch from me is not coupled to the transport. It will also
work, if you just connect two tasks inside of a main loop. The big problem I ran
into, and why there is so much patching in corba involved, is that in case of corba
the RemotePort is attached to the local task. Therefor port->getInterface()->getOwner()
returns the wrong information (local task name, not remote one). I don't see how
your proposal solves this issue.

doudou:
>  I don't see how your proposal solves this issue.

Because I assume that RTT is not the one that will fill in the information, but whatever toolchain is above it. This toolchain knows very well, at the point of connection, what are the names of everyone involved. Basically, the only thing needed is to add a field in the connection policy and a way to access it from outside.

Given that that's not your use case, though, that's a moot point.

However, storing as an application-specific format in the policy removes the need for adding all these fields that are pretty irrelevant in the general case (not all endpoints are using task and port names, if you're connecting a port to a ROS stream, the ROS topic name is informative).

In fine, I won't object to merging this if some other @orocos-toolchain/owners vet it. But I really don't see how one can easily add a way to specify the enpoint names from outside RTT with this implementation.

jmachowinski:
> Because I assume that RTT is not the one that will fill in the information, but whatever toolchain is above it.
I don't think it is a good idea to put logic, that is generally useful only in one subtoolchain.
My patch shows that this information can be gathered pretty nicely inside of RTT, without
changing the external user interface. 

I also see you point, that there is no one to one mapping for ROS connections, but one could
inject information into the system that is useful (by setting the task name to ROS and the port name to the topic). On top if this, we still have the option, to add a special field for such use cases, as you suggested.

psoetens:
I have to agree with Sylvain here. This patch may functionally work for one case, but it's *very* intrusive and not solving other cases Sylvain is mentioning.

But I also agree that this functionality is very much needed.

I would also propose to point at the ConnPolicy and/or ConnID to fill in the URL/URI/UUID data that traces back connections after creation. orogen could do that, the deployer could do that. They both have much more information than the RTT library. Why not store the used ConnPolicy (with this tracable id) in one channel element, such that the connectionmanagers on both sides can query it and other tools can use it to link ports back together, even across transports ? Maybe I'm overlooking something though...

Also worth mentioning here: there is a bug in the current ConnectionManager that it stores the default ConnPolicy and not the one used during creation. So that code is actually dead. It's the reason that the rtt_dot_service only works half-way-ish : https://github.com/orocos-toolchain/rtt_dot_service/blob/master/src/rtt_dot_service.cpp#L91 ... it always gets a default ConnPolicy, so none of the branches are taken.

jmachowinski:
There is one big point with the proposal from Sylvain, that bothers me. 
We lately switched to using pure C++ code based on Orocos API to 
create our component networks. So in our case there is a lot of

TaskProxy A;
TaskProxy B;

A.foo.connectTo(B.bar);

code around. With the solution from Sylvain, we would need to 
update all our code to get introspection functionality. Or do I
misinterpret the term 'Tooling' here ?

doudou:
The only thing you need to update is your implementation of TaskProxy. It has all the information needed to set the policy field properly without adding a lot of mumbo-jumbo in RTT itself.

jmachowinski:
> The only thing you need to update is your implementation of TaskProxy.

The TaskProxy is directly derived from the TaskContext and uses its mechanisms. To change
the connectTo call one would need to change RTT::base::OutputPortInterface::connectTo. At 
this point we would be back to the start, that at this point we need to get the information from 
the owning task.

doudou:
A.foo returns "something" that does not have to be OutputPortInterface. This "something" can have a connectTo that is overloaded, and fills the policy properly.

PR #86, opened 2015-02-09
extras: added missing initializers to one of the FileDescriptorActivity constructors and fixed compiler warnings with -Wunused-result

meyerj:
This is a follow-up and minor cleanup of pull request #78.
The missing initializers had no effect as these boolean flags are reset in start() anyway.

### DIFF
--- a/rtt/extras/FileDescriptorActivity.cpp
+++ b/rtt/extras/FileDescriptorActivity.cpp
@@ -82,6 +82,9 @@ FileDescriptorActivity::FileDescriptorActivity(int priority, RunnableInterface* 
     , m_period(0)
     , m_has_error(false)
     , m_has_timeout(false)
+    , m_break_loop(false)
+    , m_trigger(false)
+    , m_update_sets(false)
 {
     FD_ZERO(&m_fd_set);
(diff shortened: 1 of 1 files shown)


doudou:
Travis fails with an ICE on gcc 4.6. Do you guys know if it is possible to build on more recent environments ?

meyerj:
The Travis error is due to the high memory usage. Travis limits memory comsumption to 3GB per job (see http://docs.travis-ci.com/user/common-build-problems/#My-build-script-is-killed-without-any-error). The internal compiler error has nothing to do with the contents of this pull request. Sometimes it helps to simply retrigger the job.

We could try to reduce the number of threads to one (-j1). But there is also a time limit, which might become an issue without parallel builds.

The Travis VMs are based on Ubuntu 12.04 LTS Server Edition 64 bit (http://docs.travis-ci.com/user/ci-environment/#CI-environment-OS). For C++ the service supports the compilers gcc and clang, probably at the versions provided by 12.04 (not verified). However, it is possible to install a newer version of gcc from a PPA as part of the before_install hook in the Travis configuration file (https://github.com/travis-ci/travis-ci/issues/1379).

PR #88, opened 2015-03-20
Removed setting env flags to cmake variables XXXX_FLAGS_ADD

goldhoorn:
This causes a set of COMPILE_OPTIONS with CXX flags.
This causes that tlsf.c got compiled with cxx options.
Which is failing with cxx11.

The env settings of CMAKE_XXX_FLAGS is applied by default to all
packages. There is no reason to add it to custom variables.


goldhoorn:
pong?

goldhoorn:
ping?

goldhoorn:
*rebased* pong

meyerj:
I agree that the `CMAKE_XXX_FLAGS_ADD` introduced here are somehow bogus and `CMAKE_XXX_FLAGS` should be set directly. But there are various other places in rtt's cmake scripts where flags are added to one of the `*_ADD` variables, but not in the cache.

If there is no documented use case of the `CMAKE_XXX_FLAGS_ADD` variables for users or build tools I would vote for

- removing the initial setting of these variables as proposed in this pull request
- deleting all `COMPILE_FLAGS ${CMAKE_CXX_FLAGS_ADD}` and `LINK_FLAGS ${CMAKE_LD_FLAGS_ADD}` target property settings
- replacing all occurrences of

   ```cmake
SET(CMAKE_CXX_FLAGS_ADD "${CMAKE_CXX_FLAGS_ADD} ...")
```

   with

   ```cmake
SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ...")
```
   (and similar for `CMAKE_C_FLAGS_ADD` and `CMAKE_LD_FLAGS_ADD`).

goldhoorn:
I changed the CMake system as requested, and updated the PR.
The build builds fine with #94 applied with clang-3.7 and cxx11 and with normal gcc setup.
Unfortunate i cannot check the other builds (embedded), but From my point of view should work fine.

rtt/transports/mqueue/CMakeLists.txt
```diff
@@ -29,8 +29,6 @@ IF ( BUILD_STATIC )
   OUTPUT_NAME orocos-rtt-mqueue-${OROCOS_TARGET}
   CLEAN_DIRECT_OUTPUT 1
   VERSION "${RTT_VERSION}"
-  COMPILE_FLAGS "${CMAKE_CXX_FLAGS_ADD}"
-  LINK_FLAGS "${MQ_LDFLAGS} ${CMAKE_LD_FLAGS_ADD}"
```
smits: Don't we need the MQ_LDFLAGS here?

rtt/transports/mqueue/CMakeLists.txt
```diff
@@ -43,8 +41,6 @@ ENDIF( BUILD_STATIC )
   DEFINE_SYMBOL "RTT_MQ_DLL_EXPORT"
   OUTPUT_NAME orocos-rtt-mqueue-${OROCOS_TARGET}
   CLEAN_DIRECT_OUTPUT 1
-  COMPILE_FLAGS "${CMAKE_CXX_FLAGS_ADD}"
-  LINK_FLAGS "${MQ_LDFLAGS} ${CMAKE_LD_FLAGS_ADD}"
```
smits: Same here?

goldhoorn:
*this* is strance. I updated my PR (see https://github.com/orocos-toolchain/rtt/commit/d7818c220c83db5880ff8e45f597f4df97bc7cb2) but the Files Changed page did not get updated.

However, if you are fine with it now i can apply it directly to master.

goldhoorn:
...ok the page is updated, can be merged here...

PR #90, opened 2015-03-31
Second Introspection implementation

jmachowinski:
Hey, 
I did a proof of concept implementation of the concept we talked
about on the mailing list. These are the modification to rtt that are 
needed to implement the Introspection ability. Please note, that
we would also need to patch the RosPubChannelElement in order
to make it work with the ROS transport. 

Note, this is a POC, it is not intended for direct merging.

I also have a client side implementation for this, which can be found
here :
https://github.com/jmachowinski/rtt_introspection
https://github.com/jmachowinski/rtt_introspection_typekit
As it is still a proof of concept it is not tidied up and hacky....

The output of it looks like this :
Task task2
Output Ports : 
    output2 :
        Connections
            ConnInputEndpoint(0x7fe53800b7f0)->ChannelDataElement(0x7fe53800b780)->CorbaRemoteChannelElement(0x7fe53800af68)->CorbaRemoteChannelElement(IOR:010000002800000049444c3a5254542f636f7262612f4352656d6f74654368616e6e656c456c656d656e743a312e3000010000000000000064000000010102000c00000031302e3235302e332e31300024dc00000e000000feed8b1a5500007f0a000000000900000200000000000000080000000100000000545441010000001c00000001000000010001000100000001000105090101000100000009010100)->ChannelDataElement(0x7f948c0055a0)->ConnOutputEndpoint(0x7f948c004cc0)->input->task1
            ConnInputEndpoint(0x7fe53800d910)->ChannelDataElement(0x7fe53800d8a0)->CorbaRemoteChannelElement(0x7fe53800d088)->CorbaRemoteChannelElement(IOR:010000002800000049444c3a5254542f636f7262612f4352656d6f74654368616e6e656c456c656d656e743a312e3000010000000000000064000000010102000c00000031302e3235302e332e31300024dc00000e000000feed8b1a5500007f0a000000000f00000200000000000000080000000100000000545441010000001c00000001000000010001000100000001000105090101000100000009010100)->ChannelDataElement(0x7f948c0062a0)->ConnOutputEndpoint(0x7f948c005900)->input->task1

### DIFF
--- a/rtt/base/BufferLocked.hpp
+++ b/rtt/base/BufferLocked.hpp
@@ -75,7 +75,7 @@ namespace RTT
          * @param circular Set flag to true to make this buffer circular. If not circular, new values are discarded on full.
          */
         BufferLocked( size_type size, const T& initial_value = T(), bool circular = false )
-            : cap(size), buf(), mcircular(circular)
+            : cap(size), buf(), mcircular(circular), droppedSamples(0)
         {
             data_sample(initial_value);
         }
@@ -101,8 +101,12 @@ namespace RTT
         {
             os::MutexLock locker(lock);
             if ( cap == (size_type)buf.size() ) {
+                //buffer is full, we either overwrite a sample, or drop the given one
+                droppedSamples++;
                 if (!mcircular)
+                {
                     return false;
+                }
                 else
                     buf.pop_front();
             }
@@ -117,12 +121,17 @@ namespace RTT
             if (mcircular && (size_type)items.size() >= cap ) {
                 // clear out current data and reset iterator to first element we're going to take.
                 buf.clear();
+                //note the ignored samples are added below to the dropped samples.
+                droppedSamples += cap;
                 itl = items.begin() + ( items.size() - cap );
             } else if ( mcircular && (size_type)(buf.size() + items.size()) > cap) {
                 // drop excess elements from front
                 assert( (size_type)items.size() < cap );
                 while ( (size_type)(buf.size() + items.size()) > cap )
+                {
+                    droppedSamples++;
                     buf.pop_front();
+                }
                 // itl still points at first element of items.
             }
             while ( ((size_type)buf.size() != cap) && (itl != items.end()) ) {
@@ -130,10 +139,14 @@ namespace RTT
                 ++itl;
             }
             // this is in any case the number of elements taken from items.
+            size_t writtenSamples = itl - items.begin(); 
+            
             if (mcircular)
-                assert( (size_type)(itl - items.begin() ) == (size_type)items.size() );
-            return (itl - items.begin());
-
+                assert( writtenSamples == (size_type)items.size() );
+            
+            droppedSamples += items.size() - writtenSamples;
+            
+            return writtenSamples;
         }
         bool Pop( reference_t item )
         {
@@ -204,12 +217,18 @@ namespace RTT
             os::MutexLock locker(lock);
             return (size_type)buf.size() ==  cap;
         }
+        
+        size_type dropped() const
+        {
+            return droppedSamples;
+        }
     private:
         size_type cap;
         std::deque<T> buf;
         value_t lastSample;
         mutable os::Mutex lock;
         const bool mcircular;
+        size_type droppedSamples;
     };
 }}
 
--- a/rtt/TaskContext.cpp
+++ b/rtt/TaskContext.cpp
@@ -239,7 +239,7 @@ namespace RTT
     
     bool TaskContext::loadPlugin(const string& pluginPath)
     {
-        return PluginLoader::Instance()->loadLibrary(pluginPath);
+        return PluginLoader::Instance()->loadPlugin(pluginPath, "");
     }
 
 
--- a/rtt/TaskContext.cpp
+++ b/rtt/TaskContext.cpp
@@ -118,6 +118,7 @@ namespace RTT
 
         this->addOperation("trigger", &TaskContext::trigger, this, ClientThread).doc("Trigger the update method for execution in the thread of this task.\n Only succeeds if the task isRunning() and allowed by the Activity executing this task.");
         this->addOperation("loadService", &TaskContext::loadService, this, ClientThread).doc("Loads a service known to RTT into this component.").arg("service_name","The name with which the service is registered by in the PluginLoader.");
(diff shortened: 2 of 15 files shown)


jmachowinski:
We used this implementation now for a while, and it showed now problems. 
I would say it is ready for merge. Any comments ?

meyerj:
Sorry for not having commented earlier on this PR.

In general I like the idea and see that the usefulness of more advanced introspection, but the commits that add the `loadPlugin()` operation to the TaskContext interface are not related to introspection and it does not depend on the specific TaskContext instance it is called on. If the whole purpose is to make plugin loading accessible from scripting without a deployer, then it would be better-suited for addition to the GlobalService singleton instance.

I would therefore vote for reverting commits 9011cf9 and 9c9e008 on master and opening another PR that extends the PluginLoader implementation to add some of its methods as operations to the global service.

rtt/base/BufferLocked.hpp
```diff
@@ -130,10 +139,14 @@ namespace RTT
                 ++itl;
             }
             // this is in any case the number of elements taken from items.
+            size_t writtenSamples = itl - items.begin();
```
meyerj: The patch causes a bunch of new signed and unsigned integer comparison warnings to be emitted if RTT is compiled with all warnings enabled (see [Travis build #330](https://travis-ci.org/orocos-toolchain/rtt/builds/64374829), but the failure is due to resource constraints).

`size_t` -> `size_type` (== `int` as defined in BufferBase)?

meyerj:
The `BufferLockFree` implementation of the dropped samples counter is not thread-safe because the counter variable is not protected. But it is hard to do so without locking. `RTT::os::AtomicInt` might be an option.

As long as there is only one writer thread this only affects reading the concurrent reading of the counter in the `BufferLockFree::dropped()` method, but at least with #114 we also need multiple writer support for buffers.

rtt/base/BufferLocked.hpp
jmachowinski: should be fixed now, how do I get Travis to build ? 
On my local Installation I do net get the Signess Errors, even If I hard add -Wsign-compare,
-Wall is also active. One should not, I use GCC 4.8.4 while travis seems to use 4.6

jmachowinski:
Hm, do we care, that the dropped counter is not protected in the BufferLockFree case ? 
It is costly to protect it, and in the end it is 'just' a debug tool. On the other side, it
should not get triggered often, so protecting it should not be that costly...

jmachowinski:
> I would therefore vote for reverting commits 9011cf9 and 9c9e008 on master and opening another PR  that extends the PluginLoader implementation to add some of its methods as operations to the global service.

You mean move the functions into the Service class ?

rtt/base/BufferLocked.hpp
meyerj: > should be fixed now, how do I get Travis to build ? 

Sorry, we disabled the Travis CI builds a while ago, because due to the resource limits in the free plan (either in terms of memory or in terms of build time depending on the number of parallel jobs) build jobs were killed and caused too many failed builds even if there was no actual problem. I reenabled the GitHub service hook now and also switched to the relatively new [Trusty build enviroment](https://docs.travis-ci.com/user/trusty-ci-environment/) for the master and toolchain-2.9 branches (see commit https://github.com/orocos-toolchain/rtt/commit/9bd82eb25884817e9dc5c9f572a9b716ebb205f0). Travis seems to be much faster now since [an upgrade](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future) that happened in late 2015.

Your last commit to master seems to fix the sign-compare warning:
https://travis-ci.org/orocos-toolchain/rtt/builds/102463549

meyerj:
>> I would therefore vote for reverting commits 9011cf9 and 9c9e008 on master and opening another PR that extends the PluginLoader implementation to add some of its methods as operations to the global service.

> You mean move the functions into the Service class ?

No, but the constructor of the [PluginLoader](http://orocos-toolchain.github.io/devel/documentation/rtt/master/api/html/classRTT_1_1plugin_1_1PluginLoader.html) singleton could expose some of its methods as operations via the [GlobalService](http://orocos-toolchain.github.io/devel/documentation/rtt/master/api/html/classRTT_1_1internal_1_1GlobalService.html). Something like

```cpp
// untested code!!!
internal::GlobalService::Instance()->provides("plugins")->
    addOperation("loadPlugins", &PluginLoader::loadPlugins, this, ClientThread);
```

This would allow to call
```
plugins.loadPlugin("my_package", "")
```
from anywhere in scripting or in the TaskBrowser, independent of the context. But we have to be aware of potential name collisions with services, properties, ports etc. in the current component's scope. I am not sure if this also works without the `provides("plugins")` part, which would mean that the "plugins." prefix when calling could be omitted.

jmachowinski:
I don't see, how this would work. I need to call loadPlugins on foreign processes.
If I add this method to the GlobalService, it would only affect my local process ?!

Perhaps I should explain my usecase again: I use the loadPlugin call, to instrument 
the tasks at runtime, by loading the introspection service. This service adds a series
of operations that are then uses for the introspection tracking. 

This design was chosen, to keep the RTT codebase itself as light weight as possible.

meyerj:
> If I add this method to the GlobalService, it would only affect my local process?!

True. The operations in a global service cannot be called from a remote process as far as I know. I will discuss with @psoetens what is the best option to achieve this, but I would still not add it to the TaskContext core implementation. I think the "standard" way is to create a remote proxy for the Deployer component, which already provides the `import()` operation. I assume you are not using OCL's [DeploymentComponent](http://people.mech.kuleuven.be/~orocos/pub/stable/documentation/ocl/v2.x/api/html/classOCL_1_1DeploymentComponent.html)?

Would it be an option for you to add the `loadPlugin()` operation to one specific component interface (something like a custom deployer) that only exists once per process?

psoetens:
This can stay on master, but we will have to merge it with #114 , which now lives on the toolchain-2.9 branch. This step will also require minor updates on the https://github.com/jmachowinski/rtt_introspection code, but we expect it will be for the better.

We would now first release toolchain-2.9, and after the release, merge toolchain-2.9 to master.

There's also an open discussion on the loadPlugin() function, which preferably is solved by using the GlobalService, instead of a TaskContext....

PR #92, opened 2015-04-23
prevent ROS generated messages linking

ahoarau:
This removes ROS generated messages from the exported libraries that
might cause issues when auto-linking is enabled.


smits:
ptal (please take a look) @meyerj

meyerj:
Looks good to me as a workaround.

The true problem is that we assume that all exported targets are also exported libraries in [UseOROCOS-RTT.cmake:965](https://github.com/ahoarau/rtt/blob/catkin_generated_msgs_fix/UseOROCOS-RTT.cmake#L965). It would be better to keep these two clearly separated. The patch only solves this problem for the additional exported targets for message generation (which are not library targets), but not for other targets a cmake/catkin package might export.

I suggest to merge this PR and I try to come up with a cleaner solution in the close future.

PR #93, opened 2015-05-03
corba: Optionally support not emitting IORs if name service not used

snrkiwi:
A component that acts as a COBRA server, but that is not registering with the name
service, will cause an IOR to be output to the deployer console and the IOR to
be written to a file in PWD. This behaviour is not desirable in all applications.

This patch makes the behaviour optional in RTT, by providing a global option to
disable the IOR emission (defaults to OFF, to keep backwards compatibility).
When not emitting then the IOR is not written to stdout nor to file.

### DIFF
--- a/config/check_depend.cmake
+++ b/config/check_depend.cmake
@@ -44,6 +44,8 @@ endif()
 OPTION(PLUGINS_ENABLE "Enable plugins" ON)
 OPTION(PLUGINS_STD_TYPES_SUPPORT "Enable support for the std::string and std::vector<double> types in the RTT typekit & transports." ON)
     
+OPTION(ORO_NO_EMIT_CORBA_IOR "Do not emit CORBA IORs if name service not used" OFF)
+
 ###########################################################
 #                                                         #
 # Look for dependencies required by individual components #
--- a/rtt/os/targets/target-config.h.in
+++ b/rtt/os/targets/target-config.h.in
@@ -64,6 +64,8 @@
 
 #cmakedefine ORO_REMOTING
 
+#cmakedefine ORO_NO_EMIT_CORBA_IOR
+
 #cmakedefine OROBLD_DISABLE_LOGGING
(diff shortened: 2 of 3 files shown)


meyerj:
What is the purpose of these outputs?

If they are leftovers from debugging only, why not remove them completely or invert the meaning of the new cmake variable `ORO_NO_EMIT_CORBA_IOR` and switch `ORO_EMIT_CORBA_IOR` off by default?

meyerj:
Lines [TaskContextProxy.cpp:174-185](https://github.com/snrkiwi/rtt/blob/optionally-not-output-corba-ior/rtt/transports/corba/TaskContextServer.cpp#L174) and [TaskContextProxy.cpp:223](https://github.com/snrkiwi/rtt/blob/optionally-not-output-corba-ior/rtt/transports/corba/TaskContextServer.cpp#L223) should be included conditionally as well.

snrkiwi:
Because my fix is backwards compatible and does not affect existing behavior by default 
S

snrkiwi:
Ok, thanks. I'll look at those lines that I missed
S

snrkiwi:
I fixed the missing lines, thanks. BTW they're in TaskContextServer not TaskContextProxy.

Updated branch pushed.

smits:
@meyerj I dont really that Lines TaskContextProxy.cpp:174-185 and TaskContextProxy.cpp:223 should be included as well, since these are in the not using NameService branches of the code. If we want to connect to these components we need to get their IORs somehow.

snrkiwi:
Again, I believe you are talking about TaskContextServer.cpp and not TaskContextProxy.cpp. AFAICT there is no IOR emission within TaskContextProxy.cpp (as of master at 26eb8aa).

We simply want the ability for the application to prevent *all* IORs from being written to stdout or file. I believe that the as proposed patches do that. @smits is there something I'm missing?

PR #87, opened 2015-03-18
Fixed some C++11 Incompatibilities

goldhoorn:

PR #94, opened 2015-05-11
Fixed some C++11 Incompatibilities

goldhoorn:
Reopen of #87 

(i accidentally pushed rtt/master to my master's pull request which closed the PR)

### DIFF
--- a/rtt/ServiceRequester.cpp
+++ b/rtt/ServiceRequester.cpp
@@ -67,7 +67,7 @@ namespace RTT
             log(Error) << "OperationCaller with name '" + isb.getName() + "' already present." << endlog();
             return false;
         }
-        mmethods.insert(make_pair<std::string, OperationCallerBaseInvoker*> (isb.getName(), &isb));
+        mmethods.insert(make_pair(isb.getName(), &isb));
         return true;
     }
 
--- a/rtt/Handle.cpp
+++ b/rtt/Handle.cpp
@@ -81,11 +81,11 @@ namespace RTT {
     }
 
     Handle::operator bool() const {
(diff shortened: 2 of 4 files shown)


rtt/ServiceRequester.cpp
```diff
@@ -67,7 +67,7 @@ namespace RTT
             log(Error) << "OperationCaller with name '" + isb.getName() + "' already present." << endlog();
             return false;
         }
-        mmethods.insert(make_pair<std::string, OperationCallerBaseInvoker*> (isb.getName(), &isb));
+        mmethods.insert(make_pair<std::string, OperationCallerBaseInvoker*> (std::string(isb.getName()), &isb));
```
meyerj: Please change this line to
```cpp
mmethods.insert(pair<std::string, OperationCallerBaseInvoker*> (isb.getName(), &isb));
```
or
```cpp
mmethods.insert(make_pair(isb.getName(), &isb));
```
Both variants compile with and without C++11.

The whole point of `std::make_pair` is to construct pairs without having to specify the types explicitly.
See http://stackoverflow.com/questions/14623958/breaking-change-in-c11-with-make-pair-ty1-val1-const-ty2-val2.

goldhoorn: I'm not sure will try next week, the first line is the original line, i compile also with clang (not only cxx11) the problem was that getName returns a const string, and the template is a non-const string which causes the compilation to fail.

Nevertheless i will try your second proposal

goldhoorn:
Funny it did work. I updated the PR

smits:
lgtm

PR #95, opened 2015-05-17
typekit: fixed composition of type bool

meyerj:
The `BoolTypeInfo` class that overwrites the streaming operators for type bool did not inherit from `StdTypeInfo<bool>`, but directly from `TemplateTypeInfo<bool,true>`. Therefore the default composition factory methods `composeType()` and `decomposeType()` where not overwritten and composition/decomposition of PropertyBags containing bool properties failed.

Also deleted header file `StdPrimitiveTypeInfo.hpp` as it is unused and was probably replaced by StdTypeInfo.hpp.


smits:
lgtm

PR #100, opened 2015-06-11
RTT fails when using gcc5

phyrog:
RTT fails when using gcc5 (as is default on Arch Linux), but does not need to


goldhoorn:
tested on gcc 4.8 debian :+1: for merge

PR #101, opened 2015-06-12
Update toolchain-2.8 for new patch release.

smits:


rtt/typekit/BoolTypeInfo.hpp
```diff
@@ -49,10 +49,10 @@ namespace RTT
         /**
          * Write boolean as 'true' or 'false'.
          */
-        struct BoolTypeInfo: public TemplateTypeInfo<bool,true>
+        struct BoolTypeInfo: public StdTypeInfo<bool>
         {
             BoolTypeInfo() :
-                TemplateTypeInfo<bool,true> ("bool")
+                StdTypeInfo<bool> ("bool")
             {
```
smits: @meyerj Can we consider these changes ABI compatible? (I guess no-one directly uses BoolTypeInfo.hpp)

lepalom: AFAIK this change is NOT ABI compatible. See [1]. You are modifiying the signature of a struct.

[1] https://techbase.kde.org/Policies/Binary_Compatibility_Issues_With_C++

smits: I know that it's not ABI compatible, I just wonder if user code would be affected or not as this header is typically not include in user code, but is part of the rtt-typekit.

smits: I've tested this, with a component that includes rtt/typekit/Types.hpp and links with the rtt-typekit (the typical usage), and it works == ABI compatible at a user level. I verified with nm and BoolTypeInfo is nowhere referenced in the component library, only in the rtt-typekit library itself.

lepalom: I'm afraid that if someone has a component that use rtt from a package (binary) he/her will have to rebuild it, because the rtt-typekit changed ... but, if you tested it and it's ok?

Will you do a new release?

PR #98, opened 2015-05-28
boost lambda ist not working for boost 1.57 and clang 3.4, 3.5, 3.6, …

malter:

PR #103, opened 2015-06-12
boost lambda ist not working for boost 1.57 and clang 3.4, 3.5, 3.6, …

goldhoorn:
This PR superseeds https://github.com/orocos-toolchain/rtt/pull/98 because of longer vacation


snrkiwi:
Looks good to me.

smits:
lgtm

meyerj:
Is it correct that the only motivation for this patch was to  avoid `boost::lambda`? That's what I read from the commit message. Or are there any other advantages of std::bind vs. boost::bind?

I think the patch is somehow incomplete as not all occurrences of `boost::bind` have been replaced and we have a mix of Boost and C++11 functions now (e.g. [OperationInterfacePartFused.hpp:220](https://github.com/orocos-toolchain/rtt/pull/103/files#diff-7ea54099da3c95b19cb4c6cb852b43c1R220) and many others that have not been touched at all). I proposed an alternative solution using `boost::arg<i>` instead of `boost::lambda::_i` in https://github.com/orocos-toolchain/rtt/pull/196, which might also solve the problem that motivated this patch. I will try to revert this patch, replace lambdas by `boost::arg<i>` instead and check if RTT compiles with Boost 1.58 and clang. I guess the standard Ubuntu Xenial set of packages (Boost 1.58 and Clang 3.8) would reveal the original problem.

meyerj:
Please check the alternative solution in #201. I noticed that Boost Lambda was used for some expressions other than only placeholders, but those could be simply replaced by standard C++ and Boost Bind.

Issue #106, opened 2015-06-23
2.8.1-rc1 doesn't compile with xenomai target


lepalom:
Building 2.8.1-rc1 I got this error:

[ 52%] Building CXX object rtt/scripting/CMakeFiles/rtt-scripting-xenomai_plugin.dir/SendHandleAlias.cpp.o
cd "/tmp/buildd/orocos-rtt-2.8.1~rc1+dfsg/dbuild.xenomai/rtt/scripting" && /usr/bin/c++   -DOROCOS_TARGET=xenomai -DRTT_SCRIPTING_DLL_EXPORT -g -O2 -fstack-
/tmp/buildd/orocos-rtt-2.8.1~rc1+dfsg/rtt/typekit/RealTimeTypekitGlobals.cpp: In member function 'virtual bool RTT::types::RealTimeTypekitPlugin::loadGlobal
/tmp/buildd/orocos-rtt-2.8.1~rc1+dfsg/rtt/typekit/RealTimeTypekitGlobals.cpp:71:62: error: 'ORO_WAIT_ABS' was not declared in this scope
         globals->setValue( new Constant<int>("ORO_WAIT_ABS", ORO_WAIT_ABS) );
                                                              ^
/tmp/buildd/orocos-rtt-2.8.1~rc1+dfsg/rtt/typekit/RealTimeTypekitGlobals.cpp:72:62: error: 'ORO_WAIT_REL' was not declared in this scope
         globals->setValue( new Constant<int>("ORO_WAIT_REL", ORO_WAIT_REL) );
                                                              ^
rtt/typekit/CMakeFiles/rtt-typekit-xenomai_plugin.dir/build.make:57: recipe for target 'rtt/typekit/CMakeFiles/rtt-typekit-xenomai_plugin.dir/RealTimeTypeki
make[4]: *** [rtt/typekit/CMakeFiles/rtt-typekit-xenomai_plugin.dir/RealTimeTypekitGlobals.cpp.o] Error 1

PR #107, opened 2015-06-30
Added ORO_WAIT_ABS and ORO_WAIT_ABS constants for all targets

meyerj:
Fix for #106.

The constants `ORO_WAIT_REL` and `ORO_WAIT_ABS` are not available if RTT is compiled for the Xenomai target.

### DIFF
--- a/rtt/os/ecos/fosi.h
+++ b/rtt/os/ecos/fosi.h
@@ -49,6 +49,9 @@ extern "C"
 #define ORO_SCHED_RT    0
 #define ORO_SCHED_OTHER 0
 
+#define ORO_WAIT_ABS 0 /** Not supported for the ecos target */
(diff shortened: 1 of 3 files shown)


lepalom:
Why this and no add this constants to rtt/os/xenomai/fosi.h?

meyerj:
The only place where those variables are useful is the new `setWaitPeriodPolicy()` operation of the Deployer introduced in https://github.com/orocos-toolchain/ocl/pull/28. With target xenomai the corresponding member function of `RTT::os::Thread()` is a no-op and therefore the constants are not defined in C++.

But I agree that it might be better to add them to  rtt/os/xenomai/fosi.h, even if not supported, in order to keep deployment scripts compatible between targets.

I will update the patch...

PR #109, opened 2015-09-02
Can't deterministically control execution of components in a unit test that explicit depend on FileDescriptorActivity.

snrkiwi:
Any component using an FDA is explicitly coupled to the FDA activity type (i.e. it can't use Activity) as the component has to know why it was woken up, which means querying the activity object directly (for example, being woken up by a signal causes the component to process data, but being woken by a timeout causes the component to take some safety action). Because the FDA has to be queried, this creates a run-time type dependency of the component on its Activity (i.e. it *requires* an FDA as its activity as it needs to query the FDA interface - these functions (e.g. hasTimeout()) are not present in the base Activity class). Hence the component can only be instantiated with an FDA - it won't accept any other Activity type.

Our unit test cases do cycle-by-cycle checks of such component's behavior, and so they need to be able to deterministically cycle the component. But the component-under-test is explicitly coupled to, and requires an associated instance of, an FDA activity at runtime. And the FDA object itself runs a thread which is woken up by time outs or activity on its underlying file descriptor (FD) object. So for the unit test to cycle the component-under-test, it has to instantiate an FDA and associate that with the component (due to the coupling requirement), and then write to the underlying FD object *and* put the main thread (i.e. the unit test executable) to sleep to allow the FDA to pop on the CPU and wake up due to the write to the FD object. There's a very obvious race condition here - depending on how loaded the CPU(s) are, the amount of time that the main thread has to sleep to let the FDA run is non-deterministic, to say nothing of any timeout set on the FDA.

DISCARDED SOLUTIONS

You can't just swap out an FDA for something else - it must have the same interface.

You can't decouple the component-under-test from the FDA behavior (and hence use a different activity in the test case), as the component-under-test *must* know why it was woken up to be able to decide what to do.

You can't derive a new activity from FDA and instantiate the derived class, as the dynamic_cast used inside the component-under-test component will quietly provide you with the base FDA instead of the derived FDA (AFAICT by reading the C++ spec, though Peter disagrees).

You can't modify the signal pipe reading side of things to attempt determinism there, as you are still limited by the "background" thread that the FDA is running and the associated non-determinism there.

You can't use some combination of step() or trigger() within the existing FDA implementation, as you still end up with the background thread non-determinism issue, to say nothing of the timeout being involved.

We don't want to pollute our components-under-test to have them run with a FDA for hardware, and some other Activity type for unit tests. This also would prevent testing behavioral reactions to the FDA (i.e. does the component react correctly to a timeout).

PROPOSED SOLUTION

Ideally you want a method to wake the component-under-test up via a direct call from the unit test, and for the unit test to be able to explicitly indicate why the component was woken up (eg. timeout vs signal). I propose providing a new "interface" class within RTT for the FDA (e.g. RTT::FDAI), and use the existing FDA implementation as a derived class of that the FDA interface to implement the current semantics (e.g. RTT::FDA). Then you derive a new "simulated" FDA implementation (e.g. RTT::FDASimulated or similar) from RTT::FDAI that supports the needs of unit tests, and which doesn't use an actual background thread (it'd look a lot like how unit tests use a SlaveActivity, but without the master/slave relationship). Now the unit test instantiates the FDA-simulated class and associates that with the component-under-test, and that simulated FDA class provides deterministic "wake up" by cycle (and also wakeup due to data vs timeout).

### DIFF
--- a/rtt/extras/FileDescriptorActivityInterface.hpp
+++ b/rtt/extras/FileDescriptorActivityInterface.hpp
@@ -1,0 +1,177 @@
+/***************************************************************************
+
+ ***************************************************************************
+ *   This library is free software; you can redistribute it and/or         *
+ *   modify it under the terms of the GNU General Public                   *
+ *   License as published by the Free Software Foundation;                 *
+ *   version 2 of the License.                                             *
+ *                                                                         *
+ *   As a special exception, you may use this file as part of a free       *
+ *   software library without restriction.  Specifically, if other files   *
+ *   instantiate templates or use macros or inline functions from this     *
+ *   file, or you compile this file and link it with other files to        *
+ *   produce an executable, this file does not by itself cause the         *
+ *   resulting executable to be covered by the GNU General Public          *
+ *   License.  This exception does not however invalidate any other        *
+ *   reasons why the executable file might be covered by the GNU General   *
+ *   Public License.                                                       *
+ *                                                                         *
+ *   This library is distributed in the hope that it will be useful,       *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU     *
+ *   Lesser General Public License for more details.                       *
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public             *
+ *   License along with this library; if not, write to the Free Software   *
+ *   Foundation, Inc., 59 Temple Place,                                    *
+ *   Suite 330, Boston, MA  02111-1307  USA                                *
+ *                                                                         *
+ ***************************************************************************/
+
+#ifndef FILEDESCRIPTOR_ACTIVITY_INTERFACE_HPP
+#define FILEDESCRIPTOR_ACTIVITY_INTERFACE_HPP
+
+#include "../Activity.hpp"
+#include <set>
+
+namespace RTT { namespace extras {
+
+    /** An activity which is triggered by the availability of data on a set of
+     * file descriptors. step() (and hence the base::RunnableInterface's step()
+     * method) is called when data is available or when an error is encountered
+     * on the file descriptor.
+     *
+     * To use it, one must add the file descriptors to watch in the task's
+     * configureHook()
+     *
+     * <code>
+     *   FileDescriptorActivity* fd_activity =
+     *      dynamic_cast<FileDescriptorActivity*>(getActivity().get());
+     *   if (fd_activity)
+     *   {
+     *      fd_activity->watch(device_fd);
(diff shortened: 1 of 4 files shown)


doudou:
Looks good to me, :+1:

> as the dynamic_cast used inside the component-under-test component will quietly provide you with the base FDA instead of the derived FDA 

I don't see how it could be different. Only virtual methods can be successfully overriden in this context.

rtt/extras/FileDescriptorActivityInterface.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+        /** Sets the file descriptor the activity should be listening to.
+         * @arg close_on_stop { if true, the file descriptor will be closed by the
+         * activity when stop() is called. Otherwise, the file descriptor is
+         * left as-is.}
+         *
+         * This method is thread-safe, i.e. it can be called from any thread
+         *
+         * @param fd the file descriptor
+         * @param close_on_stop if true, the FD will be closed automatically when the activity is stopped
```
smits: This documented parameter is not part of the API?

rtt/extras/FileDescriptorActivityInterface.hpp
```diff
@@ -0,0 +1,177 @@
...
+         */
+        virtual void watch(int fd) = 0;
+
+        /** Removes a file descriptor from the set of watched FDs
+         *
+         * This method is thread-safe, i.e. it can be called from any thread
+         */
+        virtual void unwatch(int fd) = 0;
```
smits: Please document the argument.

rtt/extras/FileDescriptorSimulationActivity.cpp
```diff
@@ -0,0 +1,251 @@
...
+                                                                   base::RunnableInterface* _r,
+                                                                   const std::string& name) :
+    base::ActivityInterface(_r),
+    period(0),
+    running(false),
+    lastReason(base::RunnableInterface::TimeOut)
+{
+    // prevent compiler warnings
+    (void)priority;
+    (void)name;
```
smits: AFAIK, you could also do (int /*priority*/, ... , const std::string& /*name*/) instead to avoid warnings, I've got no idea which is preferred though.

snrkiwi:
Doxygen comments fixed, per Ruben's request.

I also removed some references to the RunnableInterface::WorkReason work we've been doing on a side branch. This effectively makes the FileDescriptorSimulationActivity useless for now, but when the RunnableInterface work is integrated then we can easily fix this.

Please review and merge.

smits:
lgtm, @meyerj, @psoetens could you take a look as well please.

meyerj:
+1

snrkiwi:
Ping ….???

PR #112, opened 2015-09-14
tests: fixed typos in corba and mqueue ipc tests that caused segfaults

meyerj:
This is a regression from 180e6905c43cb23a76f588783179e14ff5071caf that causes the corba-ipc-server and corba-mqueue-ipc-server processes to segfault if real-time memory allocation is enabled (the default if TLSF is available).


smits:
lgtm

PR #113, opened 2015-09-15
scripting: added missing zero-pointer checks in StateMachine::executePending()

meyerj:
This is a follow-up of 6537ae7736ba887c25642f76ac96081d04039c5e that was merged in #64. Some pointer checks were missing.

This patch is completely untested! I don't even know at the moment why the current pointer can be NULL in this method, but this is obviously the reason for the previous PR.


smits:
lgtm

PR #114, opened 2015-10-27
New dataflow semantics with shared input and output buffers

meyerj:
This pull request implements new types of port connections with different semantics. The main addition is the support of shared input and/ or output buffers, where either multiple output ports write to the same input buffer, multiple input ports read from the same output buffer, or both. The latter case is basically reinstalling the connection semantics of RTT v1, where a connection was not much more than a shared data object or buffer.

The motivation and some implementation details of this patch are outlined in this design document:
https://docs.google.com/document/d/1zDnPPz4SiCVvfEFxYFUZBcVbXKj33o4KaotpCtXM4E0/pub

An earlier version of the document has been discussed on the orocos-dev mailing list some weeks ago:
http://www.orocos.org/forum/rtt/rtt-dev/updated-dataflow-semantics-rtt

The primary conclusion from this thread was that the default semantics should not be changed within RTT v2 and the new options have to be selected by the user explicitly. So this is what this patch (hopefully) achieves: Without explicitly selecting the new read or write policies in the ConnPolicy the resulting connection architecture is fully backwards compatible to previous versions of RTT v2. However, the new underlying channel API made it necessary to also adapt some classes which might be used in user code, but we expect that no code changes will be required for most user applications. Custom transport implementations (like for example the ROS transport implemented in [rtt_roscomm](https://github.com/orocos/rtt_ros_integration/tree/HEAD/rtt_roscomm) will certainly require an update to make them work 

In order to simplify the review process, we split the patch into multiple parts that can be merged independently and for each of them all existing test cases pass successfully (only verified with target gnulinux so far). The [updated-dataflow-semantics-rebased](https://github.com/orocos-toolchain/rtt/tree/updated-dataflow-semantics-rebased) will be continuously rebased on top of the current master once partial pull requests have been merged in.
- [x] Added version comparison macro: #115
- [x] Use const reference for templated streaming operator in RTT::Logger: #116
- [x] Adapt the data object and buffer implementations to support multiple writer and/or reader threads and fixed a race condition that affects the returned FlowStatus in ChannelDataElement: #117
- [x] Added new field to the ConnPolicy struct: #118
- [x] Added an implementation of a shared ownership mutex: #119

The remaining patches actually implement the new data flow policies and add new tests. Please look at the commit messages for further details, especially for the biggest patch afe68a89ea8fb5d4c631411cf2af56573610573a.

There is also a new test, the [dataflow_performance_test](https://github.com/orocos-toolchain/rtt/blob/updated-dataflow-semantics-rebased/tests/dataflow_performance_test.cpp), which evaluates the performance of write and read operations in different multi-threaded writer/reader scenarios and with various connection policies. I will add some results as an appendix to the design document soon.


meyerj:
The doxygen API documentation for this branch can be browsed here:
http://orocos-toolchain.github.io/devel/documentation/rtt/updated-dataflow-semantics-rebased/api/html/

doudou:
Hey. I don't have time to do a thorough code review, so I mainly checked first onion skins (public API and a few other "close-to-public"). Looks fine to me for all intent and purposes.

My understanding so far is that all custom transport implementations (e.g. ROS) will have to be updated, is that right ?

One test that was missing in the RTT codebase (and that I shamelessly implemented in orocos.rb when I fixed the bug, shame on me) is having connection/disconnections concurrently to read/write. Given that you touch the channel implementation, I advise you to implement such a stress test and make sure it passes.

meyerj:
Hi Sylvain,

thanks for your feedback!

Hey. I don't have time to do a thorough code review, so I mainly checked
> first onion skins (public API and a few other "close-to-public"). Looks
> fine to me for all intent and purposes.
>
> My understanding so far is that all custom transport implementations (e.g.
> ROS) will have to be updated, is that right ?
>
Yes, unfortunately. What is conceptually different now is that transports
should not install buffers themselves anymore, unless they add their own
dispatcher thread to handle it. For push connections the writing channel
element has to implement write() and the reading channel element simply
forwards new samples to its output without caching (e.g. from a roscpp
callback). For pull connections the writing side has to call read() on its
input and the reading side must implement the read() method. If the
transport only supports either push or pull, stream creation should fail in
the other case. Note that the push/pull flag can have different values on
the sending and receiving side of the stream, which makes sense if the
underlying transport implementation can do the caching (e.g. in the network
stack). This certainly has to be documented somehow. Is there any existing
documentation about custom transports. Release notes?

I already updated the rtt_roscomm package and pushed my changes to the
updated-dataflow-semantics branch of rtt_ros_integration:
https://github.com/orocos/rtt_ros_integration/pull/48


> One test that was missing in the RTT codebase (and that I shamelessly
> implemented in orocos.rb when I fixed the bug, shame on me) is having
> connection/disconnections concurrently to read/write. Given that you touch
> the channel implementation, I advise you to implement such a stress test
> and make sure it passes.
>
I agree that this would be nice to have. I will try to add such a test
during the next days.

meyerj:
I pushed some performance test results to this GitHub repo: https://github.com/meyerj/rtt_dataflow_performance_test

These results have been generated with the latest [dataflow_performance_test.cpp](https://github.com/orocos-toolchain/rtt/blob/f0d8953f1c17d82a5e25757de8611b2f7edfdab0/tests/dataflow_performance_test.cpp)@f0d8953 compiled against this branch, the current RTT v2 master and RTT v1.12.1.

meyerj:
>> One test that was missing in the RTT codebase (and that I shamelessly implemented in orocos.rb when I fixed the bug, shame on me) is having  connection/disconnections concurrently to read/write. Given that you touch the channel implementation, I advise you to implement such a stress test  and make sure it passes.

> I agree that this would be nice to have. I will try to add such a test
during the next days.

@doudou  Is this more or less what you expected? #120

The new test already revealed a small bug that made it impossible to connect a second reader to a `WriteShared` output port with the `init` flag set (fixed in b8a5f00).

meyerj:
merged into toolchain-2.9: 6707ddfa16a0f5604fe069f9160ece47f0b634c5

PR #117, opened 2015-10-27
Fixed FlowStatus race condition in ChannelDataElement and fixed BufferLockFree implementation for the circular buffer case

meyerj:
- Moved storage of the `FlowStatus` to the `DataObjectInterface` or `BufferInterface` implementations instead as unprotected member variables of `ChannelDataElement`
- Decide whether to use an `AtomicMWMRQueue` (renamed from `AtomicQueue`) or an `AtomicMWSRQueue` in `BufferLockFree` depending on whether the buffer is circular or must support multiple reader threads.
  For a circular buffer also the writer thread may consume elements if the buffer is full, so the queue has to support multiple readers. This was broken in previous versions of RTT v2 and could lead to buffer corruption on concurrent writer/reader access.
- Added a boolean return value to `DataObjectInterface::Set()` in order to notify the caller whether setting the data object succeeded or not. It can only fail for `DataObjectLockFree` if too many readers try to access it concurrently.
- Allow to construct uninitialized data objects or buffers and added a boolean argument to `data_sample(...)` to enforce reinitialization or not
- Added base class `base::DataObjectBase` for all data object types, similar to existing `base::BufferBase`
- Added new structs `base::BufferBase::Options` and `base::DataObjectBase::Options` to pass options to the constructors of buffer and data objects to avoid ambiguity
- Store the ConnPolicy used to construct the data object or buffer in `ChannelDataElement` or `ChannelBufferElement`, respectively
- Added multi-threaded tests for class `TsPool`, data objects and buffers

### DIFF
--- a/rtt/base/DataObjectLockFree.hpp
+++ b/rtt/base/DataObjectLockFree.hpp
@@ -43,6 +43,7 @@
 #include "DataObjectInterface.hpp"
 #include "../Logger.hpp"
 #include "../types/Types.hpp"
+#include "../internal/DataSourceTypeInfo.hpp"
 
 namespace RTT
 { namespace base {
@@ -80,8 +81,11 @@ namespace RTT
         : public DataObjectInterface<T>
     {
     public:
+        typedef typename DataObjectInterface<T>::value_t value_t;
+        typedef typename DataObjectInterface<T>::reference_t reference_t;
+        typedef typename DataObjectInterface<T>::param_t param_t;
+
         typedef typename DataObjectBase::Options Options;
-        typedef typename DataObjectInterface<T>::DataType DataType;
 
         /**
          * @brief The maximum number of threads.
@@ -107,7 +111,7 @@ namespace RTT
             {
                 oro_atomic_set(&counter, 0);
             }
-            DataType data;
+            value_t data;
             FlowStatus status;
             mutable oro_atomic_t counter;
             DataBuf* next;
@@ -150,7 +154,7 @@ namespace RTT
          * @param max_threads The maximum number of threads accessing this DataObject.
          * @note You have to specify the maximum number of threads explicitly
          */
-        DataObjectLockFree( const T& initial_value, const Options &options = Options() )
+        DataObjectLockFree( param_t initial_value, const Options &options = Options() )
             : MAX_THREADS(options.max_threads()), BUF_LEN( options.max_threads() + 2),
               read_ptr(0),
               write_ptr(0),
@@ -169,12 +173,12 @@ namespace RTT
         /**
          * Get a copy of the data.
          * This method will allocate memory twice if data is not a value type.
-         * Use Get(DataType&) for the non-allocating version.
+         * Use Get(reference_t) for the non-allocating version.
          *
          * @return A copy of the data.
          */
-        virtual DataType Get() const {
-            DataType cache = DataType();
+        virtual value_t Get() const {
+            value_t cache = value_t();
             Get(cache);
             return cache;
         }
@@ -189,7 +193,7 @@ namespace RTT
          *                      has not been updated since the last call.
          * @param copy_sample   If true, copy the data unconditionally.
          */
-        virtual FlowStatus Get( DataType& pull, bool copy_old_data, bool copy_sample ) const
+        virtual FlowStatus Get( reference_t pull, bool copy_old_data, bool copy_sample ) const
         {
             if (!initialized && !copy_sample) {
                 return NoData;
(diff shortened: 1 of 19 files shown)


rtt/base/DataObjectLockFree.hpp
```diff
@@ -198,34 +237,77 @@ namespace RTT
              * not increment the counter on write_ptr+1). Hence, no
              * locking is needed.
              */
+
+            if (!initialized) {
+                types::TypeInfo *ti = types::Types()->getTypeInfo<DataType>();
```
doudou: .... I would suggest to at least add a way for the TypeInfo object to tell that the type is RT-safe by itself.

And a lot of Rock users really don't care that much about RT-safety. Having logs and output filled with errors that you really don't care about is really not that great. Could we at least demote this message to Info ? Or have a way to disable it ?

rtt/base/DataObjectLockFree.hpp
```diff
@@ -198,34 +237,77 @@ namespace RTT
...
-            data[BUF_LEN-1].data = sample;
-            data[BUF_LEN-1].next = &data[0];
+        }
+
+        // This is actually a copy of Get(), but it only sets the status to NoData once a valid buffer has been found.
+        // Subsequent read() calls will read from the same buffer and will return NoData until a new sample has been written.
+        virtual void clear() {
+            if (!initialized) return;
+
+            PtrType reading;
```
doudou: Wouldn't it make sense to factor out this (complex) piece of logic into a acquireReadPointer or something like that ?

rtt/internal/ChannelBufferElement.hpp
```diff
@@ -129,24 +133,31 @@ namespace RTT { namespace internal {
...
+                buffer->Release(last_sample_p);
+            last_sample_p = 0;
             buffer->clear();
             base::ChannelElement<T>::clear();
         }
 
         virtual bool data_sample(param_t sample)
         {
-            buffer->data_sample(sample);
+            if (!buffer->data_sample(sample)) return false;
```
doudou: Is it on purpose that the call to the base class is bypassed if `buffer->data_sample(sample)` returns false ? I kind of fail to really see what possible consequences this could have...

rtt/internal/ChannelBufferElement.hpp
```diff
@@ -129,24 +133,31 @@ namespace RTT { namespace internal {
...
 
         virtual T data_sample()
         {
             return buffer->data_sample();
         }
-        
+
+        /** Returns a pointer to the ConnPolicy that has been used to construct the underlying buffer.
+        */
+        virtual const ConnPolicy* getConnPolicy() const
```
doudou: Why a pointer (vs plain object or reference) ?

rtt/base/BufferBase.cpp
```diff
@@ -38,7 +38,43 @@
...
+    : circular_(circular)
+    , max_threads_(2)
+    , multiple_writers_(false)
+    , multiple_readers_(false)
+{}
+
+BufferBase::Options::Options(const ConnPolicy &policy)
+    : circular_(policy.type == ConnPolicy::CIRCULAR_BUFFER)
+    , max_threads_(2)
+//    , multiple_writers_(policy.read_policy == ReadShared)
```
doudou: Do you really mean to commit these commented-out code ?

meyerj: Yes: This PR must be considered as a part of the bigger data flow refactoring in #114. The complete patch sets the `multiple_writers_` and `multiple_readers_` flags from the `read_policy`/`write_policy` fields of the connection policy. I only commented the initializers here because the new fields are not yet available without #118.

rtt/base/DataObjectLockFree.hpp
meyerj: You are right that not having set a data sample is not an error for many types and the error message should be degraded to warning or info level. On the other side I never saw this triggering for the normal usage of the lock-free data object in port connections, at least not for local connections. The OutputPort always sets a sample on the connection in [OutputPort.hpp:122](https://github.com/meyerj/rtt/blob/8d7c478d605a95271acefaa753703dd3fa548056/rtt/OutputPort.hpp#L122). I would have to verify if this is also the case for remote connections and/or streams.

Did you actually see this message while testing this branch with Rock?

rtt/internal/ChannelBufferElement.hpp
meyerj: With #114 this method implements [ChannelElementBase::getConnPolicy()](https://github.com/orocos-toolchain/rtt/pull/114/files#diff-51b13774fd6d17869ba8eb2ba6588f7aR238) and the default implementation in [ChannelInterface.cpp:208](https://github.com/orocos-toolchain/rtt/pull/114/files#diff-dff412a478239365fa7700cf3231f045R208) returns a null pointer. Not every channel element stores its connection policy. Only special implementations like data or buffer elements do.

The predecessor of this method was [ChannelElementBase::getConnID()](https://github.com/orocos-toolchain/rtt/pull/114/files#diff-51b13774fd6d17869ba8eb2ba6588f7aL178), which also returned a pointer that could be null.

PR #118, opened 2015-10-27
Updated dataflow semantics ConnPolicy

meyerj:

The new fields added to the `ConnPolicy` struct are:
- `read_policy` (of enum type `ReadPolicy`)
- `write_policy` (of enum type `WritePolicy`)
- `mandatory` (of type bool)
- `max_threads` (of type int).

We also added a global ConnPolicy instance which serves as a template for new default-constructed instances. It can be accessed from scripting through the GlobalsRepository as `DefaultConnPolicy`.

Please check the source code documentation in [ConnPolicy.hpp](https://github.com/meyerj/rtt/blob/updated-dataflow-semantics-connpolicy/rtt/ConnPolicy.hpp) or the [design document](https://docs.google.com/document/d/1zDnPPz4SiCVvfEFxYFUZBcVbXKj33o4KaotpCtXM4E0/pub) for more details.


meyerj:
In the meantime we had some discussions about the naming of the new policy fields and the corresponding constants. Even during internal discussions the current naming scheme leads to some confusion, especially because it is not clear to the user what exactly is shared in a `ReadShared` (multiple writers write to the same input buffer) or `WriteShared` connection (multiple readers read from the same output buffer).

We therefore consider to unify the `read_policy` and `write_policy` fields into a single field called the `buffer_policy` with four possible values:

* `PerConnection <==> (read_policy == ReadUnordered) && (write_policy == WritePrivate)`
   Buffers (or data objects) will be installed per connection (per pair of output and input port/stream).  Input ports with multiple connections first try to read from the last read channel first, then poll all connections in the order they have been made (which in practice means there are no guarantees on the order if multiple writers write concurrently). This is the old and new default buffer policy. 

* `PerInputPort <==> (read_policy == ReadShared) && (write_policy == WritePrivate)`
   Every input port has a single input buffer (or data object) and all connected output ports/streams will write to the same buffer. This policy requires that `pull == false` for remote connections or input streams.

* `PerOutputPort <==> (read_policy == ReadUnordered) && (write_policy == WriteShared)`
   Every output has a single output buffer (or data object) and all connected readers "consume" elements from this buffer. Exactly one connected input port or output stream will see every written sample as `NewData`. This policy requires that `pull == true` for remote connections or output streams.

* `Shared <==> (read_policy == ReadShared) && (write_policy == WriteShared)`
   The buffer (or data object) is shared between all connected input and output ports. It can have an arbitrary number of writers and readers. Ports can be connected to an existing shared connection instance either by connecting it to a port that is already part of the shared connection group, or with a special [createConnection()](https://github.com/orocos-toolchain/rtt/blob/updated-dataflow-semantics-rebased/rtt/base/PortInterface.hpp#L180) method. This was the default buffer policy in the days of RTT v1.

What do you think?

doudou:
:+1: looks less confusing indeed.

psoetens:
:+1:

smits:
:+1:

PR #119, opened 2015-10-27
Added new mutex class SharedMutex which implements a shared lock ownership mode

meyerj:

The new mutex type implements a shared ownership lock that allows concurrent access to shared data by multiple readers and exclusive access for writers. This lock type is used to protect the input and output pointers lists in the new `ChannelElementBase` implementations.

The concept of a shared mutex is targeted to be added to the C++ standard library sooner or later as [std::shared_mutex](http://en.cppreference.com/w/cpp/thread/shared_mutex) and provided by the Boost thread library as [boost::shared_mutex](http://www.boost.org/doc/libs/1_54_0/doc/html/thread/synchronization.html#thread.synchronization.locks.shared_lock).

### DIFF
--- a/rtt/os/CAS.hpp
+++ b/rtt/os/CAS.hpp
@@ -59,7 +59,9 @@ namespace RTT
      * Overload of Compare And Swap for oro_atomic_t.
      */
     inline bool CAS( volatile oro_atomic_t* addr, const int expected, const int value) {
-        return expected == oro_cmpxchg(&oro_atomic_read(addr), expected, value);
+        // assume that oro_atomic_read(addr) returns a reference
+        volatile int &ref = oro_atomic_read(addr);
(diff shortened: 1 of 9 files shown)


meyerj:
The `RTT::os::SharedMutex` has been rewritten and the mutex that has been used to protect the state information has been removed, which could be problematic and cause priority inversion issues. Non-exclusive locks on shared mutexes are used extensively in the new dataflow implementation introduced in https://github.com/orocos-toolchain/rtt/pull/114 to protect the channel pipeline from concurrent connects and disconnects, but in most cases they are only ever locked by a single thread. On the other hand a comparison with the previous implementation did not show any performance improvements using the `gnulinux` target, probably because pthread mutexes are already optimized for the case that there is no congestion.

PR #120, opened 2015-11-06
Added test suite to test connecting and disconnecting ports concurrently

meyerj:
These new tests run 4 threads each. Two of them are continuously writing to and reading from an OutputPort/InputPort pair, while the other two connect and disconnect an additional OutputPort and an additional InputPort to these ports, respectively.

The four test cases cover the four different connection layouts introduced with the updated-dataflow-semantics branch (see #114).

### DIFF
--- a/rtt/base/PortInterface.cpp
+++ b/rtt/base/PortInterface.cpp
@@ -64,6 +64,10 @@ PortInterface& PortInterface::doc(const std::string& desc) {
     return *this;
 }
 
+bool PortInterface::connectedTo(PortInterface* port) {
+    return cmanager.connectedTo(port);
+}
+
 bool PortInterface::isLocal() const
 { return serverProtocol() == 0; }
 int PortInterface::serverProtocol() const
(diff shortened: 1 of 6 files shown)


doudou:
:+1:

Wouldn't it make sense to add a connector thread between the two base ports as well ?

meyerj:
> Wouldn't it make sense to add a connector thread between the two base ports as well ?

Done: 84be7cb9c0c6c46ddce05ce1c67b51ec9f5d0259

I had to convert some raw pointers to shared pointers to prevent segfaults during massive concurrent port write/read/connect/disconnect operations (8cd24d0a7c245780bae4bc8df67aafe72b9bc761). I am not sure if concurrent connect/disconnect and read/write method calls on the same port object have been supported before. At least this case was not covered by a test.

doudou:
> At least this case was not covered by a test.

It was working, and it was supported by a test. Unfortunately I got lazy and added that test to orocos.rb, not RTT

PR #123, opened 2015-12-15
CORBA efficiency patches

meyerj:
This PR is the result of some work we did last year to optimize the CORBA inspection and transport, especially for systems where the network is affected by transport delay and packet loss. Actually I already wanted to open a PR to merge into master since a long time, but now #122 reminded me that I actually never did that. This branch has already been used in some bigger deployments and I would consider it as ready to be merged.

The existence of this patch has already been mentioned in a side note during the update dataflow semantics discussion on the mailing list: http://www.orocos.org/forum/rtt/rtt-dev/updated-dataflow-semantics-rtt#comment-36662
### Major updates

Most changes in the CORBA transport affect the way a TaskContextProxy synchronizes its objects (ports, operations, properties, services etc.) with the remote server. The number of CORBA calls going over the network has been greatly reduced. In fact, the proxy requests one big TaskContextDescription object in a single call now instead of enumerating all task objects subsequently.

Furthermore ~~write() and~~ (optional, reverted in https://github.com/orocos-toolchain/rtt/pull/151) remoteSignal() calls for port connections via a CORBA transport have been converted to oneway calls, which means that the writer (the CORBA dispatcher of the writing TaskContext) is not blocking anymore (see also #122). The signalling can be disabled completely if appropriate (see section Optional patches).

### Caveats

The IDL is not compatible to previous RTT releases, so both sides of a connection have to updated synchronously.

### Optional patches

The following patches are only enabled on demand (at compile time of RTT) as they can change the behavior of the application, beyond pure optimization. They can be enabled by setting the corresponding Cmake variables to ON. Default state is OFF.

#### PLUGINS_CORBA_PORTS_DISABLE_SIGNAL

When a port is connected to a remote port with the pull flag of the ConnPolicy set to true, the writing side normally signals the reader to notify him about new data. If `PLUGINS_CORBA_PORTS_DISABLE_SIGNAL` is enabled, the `remoteSignal()` call is skipped. This is only for saving bandwidth if data is written periodically anyway. The subsequent read() call to fetch the data blocks the reader, so pulling connections are not well suited for robot to GUI connections anyway unless the GUI reads only on demand or there is a separate thread that periodically reads all input ports.

Note that adding input ports which are only connected through a remote connection to a component as an event port will not trigger the component if remote signalling is disabled.

#### PLUGINS_CORBA_NO_CHECK_OPERATIONS

Normally the remote side checks the validity of the signature whenever it produces a handle for a remote operation call (or send), so for every OperationCaller connection in C++ or every call to a remote operation in scripting. The PLUGINS_CORBA_NO_CHECK_OPERATIONS cmake flag will disable this check, which means that wrong signatures cannot be detected while the OperationCaller is connected (usually in the configureHook or startHook) or when a script is loaded. Calling or sending an operation with a wrong signature will results in an exception during runtime instead.
Omitting the checkOperation() calls greatly reduces the setup time for the GUI (approx. by factor 2).

#### PLUGINS_CORBA_SEND_ONEWAY_OPERATION

This cmake switch enables oneway sending of operations which have a collect arity of zero (return void and have no reference arguments). The advantage is that the send() call does not block until the CSendHandle is returned. On the other side, without the handle it is impossible to check whether the operation has finished or if an exception occured. Exceptions or cases where an operation has been send with the wrong signature (if PLUGINS_CORBA_NO_CHECK_OPERATIONS is also active) at least prints an error log message at the server side.
Calling operations is not affected by this patch and always blocks until the operation finishes and the result is returned by the remote side.

#### PLUGINS_CORBA_PORTS_WRITE_ONEWAY

Although port writes could be communicated through oneway method calls on the CORBA layer for efficiency, the original patch (https://github.com/orocos-toolchain/rtt/pull/123/commits/08b77c4ce6c20cf306c4d86767e85528f3dfa31a) which applied this change unconditionally broke the guaranty that input ports read the samples in the same order as they have been written to the output port. Due to multi-threading in the CORBA implementation (depending on its implementation and its configuration) this guaranty did not hold anymore, which caused issues (discussion starting in https://github.com/orocos-toolchain/rtt/pull/123#issuecomment-183443342) and occasional unit test failures.

Therefore this part has been reverted in https://github.com/orocos-toolchain/rtt/pull/151 and replaced by an optional build-time configuration variable PLUGINS_CORBA_PORTS_WRITE_ONEWAY.


meyerj:
rebased on master

rtt/transports/corba/OperationInterface.idl
```diff
@@ -125,17 +137,18 @@ module RTT
      */
     interface COperationInterface
     {
-	typedef sequence<string> COperationList;
+      typedef sequence<string> COperationList;
```
planthaber: Had to adapt rock's ruby bindings to this PR for my test in https://github.com/orocos-toolchain/rtt/pull/122

Hab to switch to COperationDescriptions from COperationList and found that the COperationList typedef here is still defined but not used anymore.

doudou:
Since it's incompatible for e.g. the Rock bindings, would it be possible to have a way to test for the change at the preprocessor level ? That would allow to be compatible with both the new and old interface.

meyerj:
I started to create [toolchain-2.9](https://github.com/orocos-toolchain/rtt/tree/toolchain-2.9) branches which have the corba-efficiency-patches (#123), the update-dataflow-semantics (#114) and also the new `RTT_VERSION_GTE` preprocessor macro (#115) already integrated. With this branch you could compile [tools-orocosrb](https://github.com/rock-core/tools-orocosrb) conditionally:
```cpp
#if RTT_VERSION_GTE(2,9,0)
// ...
#endif
```
There might be more small API changes that require adaptations in the Ruby bindings for rtt 2.9.
Or do you had something else in mind with "a way to test for the change at the preprocessor level"?

I am currently working on the Toolchain 2.9 release which has most of the open PRs with API breaking changes merged in. Peter already wrote down some release notes (https://github.com/orocos-toolchain/orocos_toolchain/pull/12). I wanted to ask the orocos-dev mailing list for comments during the next days, especially what branch/tag of typelib and orogen should be considered as stable. For now I am using the rock-15.05 tag present in most of the repos.

rtt/transports/corba/OperationInterface.idl
meyerj: Indeed, `COperationList` should be removed.

doudou:
The problem with orogen is that you still have a pretty heavily forked version of it. The major difference it will make for you between rock-15.05 and master is that master is "ready" for clang (not meaning that it works, but at least most of the ground work is there)

meyerj:
From https://github.com/orocos-toolchain/rtt/pull/122#issuecomment-165915907:
> It should be noted that with the merged branches corba-efficiency-patches (#123) and updated-dataflow-semantics-rebased (#114) in toolchain-2.9 writes are only send as one-way calls if the connection is not mandatory (see new flag mandatory in ConnPolicy.hpp). The differentiation happens in RemoteChannelElement.hpp:351.

After I wrote this I noticed that because all writes through CORBA are buffered and forwarded by the CorbaDispatcher thread, the `WriteStatus` return value as returned by the `RemoteChannelElement` is invisible to the caller anyway. In that case we could also stick to one-way writes in all cases. I updated the merge commit in 3643898159eee5aa59f961dc90c5b54d240e09e4.

On the other hand it must be clearly documented that failures during transport (e.g. packet loss) or at the remote side (e.g. full input buffer) are not reflected in the `WriteStatus` returned to the original writer.

meyerj:
Just found another caveat when writes are oneway: There is no guaranty on the order of arrival of write operations anymore:

```
Task A: 1.588 [ Debug  ][Thread] RemoteChannelElement(0x111f8d0): write(6.33) -> CORBA
Task A: 1.588 [ Debug  ][Thread] RemoteChannelElement(0x111f8d0): write(6.33) finished.
Task A: 1.588 [ Debug  ][Thread] RemoteChannelElement(0x111f8d0): write(3.33) -> CORBA
Task A: 1.588 [ Debug  ][Thread] RemoteChannelElement(0x111f8d0): write(3.33) finished.
Task B: 380.228 [ Debug  ][SendHandleC] RemoteChannelElement(0x7fb89c1b19a0): CORBA -> write(3.33)
Task B: 380.228 [ Debug  ][SendHandleC] RemoteChannelElement(0x7fb89c1b19a0): CORBA -> write(6.33)
Task B: 380.228 [ Debug  ][SendHandleC] Received 3.33
Task B: 380.228 [ Debug  ][SendHandleC] Received 6.33
```

doudou:
Indeed: http://www.omniorb-support.com/pipermail/omniorb-list/2005-April/026538.html

meyerj:
> Since it's incompatible for e.g. the Rock bindings, would it be possible to have a way to test for the change at the preprocessor level ? That would allow to be compatible with both the new and old interface.

@doudou What exactly would be incompatible? Can you point me to some code? At the Port or TaskContext(Proxy) API level you should not see a difference. Of course the CORBA IDL and the RemoteChannelElement API changed.

Anyway, you can check the version of RTT with a macro introduced in #115, so that if this PR would be applied to version 2.9 only, you can write code like
```cpp
#include <rtt/rtt-config.h>

#ifdef RTT_VERSION_GTE
#if RTT_VERSION_GTE(2,9,0)
    // with corba-efficiency-patches
#endif
#else
    // without corba-efficiency-patches
#endif
```
There are no preprocessor macros that allow conditional compilation directly based on the availability of certain RTT features.

doudou:
> What exactly would be incompatible?

The Ruby bindings access the CORBA API directly. [This has been handled](https://github.com/rock-core/tools-orocosrb/commit/4598a518393ccdb938581f3d1f944625fe17f9b1) with the RTT_VERSION_GTE macro already, so that's not an issue anymore.

doudou:
Anyways, the lack of an ordering guarantee is a deal-breaker

planthaber:
I agree, but loosing complete messages in distributed systems is also a severe issue:

As  you are using rock, you can easily try the minimal example I posted in my original PR:
https://github.com/orocos-toolchain/rtt/pull/122

But in the end, omitting messages on network congestion is the only way to obtain real-time.

I'm not deep enough into the code, but non-blocking sends or multi-threaded sends on a single port would at least fix the issue of loosing all local sends, when the reply of the remote component takes too long.

planthaber:
Perhaps, setting the maxServerThreadPerConnection = 1 parameter in the CORBA::ORB_init() (ApplicationServer.cpp:62) call is an option?

http://omniorb.sourceforge.net/omni41/omniORB/omniORB004.html

meyerj:
> Perhaps, setting the maxServerThreadPerConnection = 1 parameter in the CORBA::ORB_init() (ApplicationServer.cpp:62) call is an option?

I was also thinking about that option when I read the post that @doudou referenced above. I tried it, but, unfortunately, even if the `maxServerThreadPerConnection` parameter is set to one, it happens occasionally that the corba-ipc-test fails because of two reordered write calls. My subjective impressions is that the likelihood of the failure went down, but still, it can happen. From increasing the traceLevel I learned that the client had opened a second GIOP connection to the same server (different TCP source port, same destination port). Aha! So I also set the client side parameter `maxGIOPConnectionPerServer` to 1 (default is 5), but unfortunately this causes a dead-lock during the `testRemoteOperationCallerCallback` test case.

Another related chapter from the omniORB documentation:
http://omniorb.sourceforge.net/omni41/omniORB/omniORB008.html

What is not clear to me is what exactly is a "server" for CORBA. The whole process? A TaskContext? Every RemoteChannelElement?

A quick workaround would be to either add a `reliability` flag to the connection policy which explicitly allows to use unreliable transports (like CORBA oneway) and stick to normal two-way calls in the default case. This option is orthogonal to existing fields and could also be implemented by other transports (e.g. [TCPROS](http://wiki.ros.org/ROS/TCPROS) vs. [UDPROS](http://wiki.ros.org/ROS/UDPROS)). The alternative would be to make it either an application-wide run-time or a compile-time option. Thoughts?

In case of the [remoteSignal() call use by pull connections](https://github.com/orocos-toolchain/rtt/pull/123/files#diff-d4440bb50044922398d16b16cfb915cdR82) it should be fine to keep the oneway keyword because the order of arrival does not matter if multiple callers signal the same channel concurrently.

planthaber:
IMHO a setting it globally is not an option and not necessary. Local connection really work very fine when waiting for replies. The possibility of congestion on the loopback device is quite low.

Where is is higher on , let's say, WLAN connections, but those are most likely telemetry connections for some GUI where unordered of lost messages are not that severe. 

So GUIs could use that flag for inbound connections, that would solve the issue of the task waiting for the reply and skipping local messages.

Nevertheless, for inexperienced users this issue might be hard to debug. It took long enough for us to figure out why the Robot behaves "stange" when we connected Rocks TaskInspector GUI remotely (because local messages, like single lines of a laser scan, where lost).

doudou:
IMO the right solution to isolate connections is to partition the network by allowing to select different CORBA dispatchers when creating connections (e.g. one dispatcher per host, or one per "target system", or ...). Doing it at the method-call level is definitely not a solution as CORBA is not meant for that.

Given the reordering issue, I am definitely NOT for this PR - in no circumstances. Limiting the number of threads per connection could very well make the performance situation worse, not better, as omniORB would not be allowed to spawn new threads to serve calls in parallel.

meyerj:
> IMO the right solution to isolate connections is to partition the network by allowing to select different CORBA dispatchers when creating connections (e.g. one dispatcher per host, or one per "target system", or ...). Doing it at the method-call level is definitely not a solution as CORBA is not meant for that.

Reducing the number of dispatchers in a process would not solve the problem. Dispatchers are at the client side, but the problem is that even calls from the same dispatcher in the same thread can be handled by different threads at the server side without setting the `maxServerThreadPerConnection` parameter to 1. It would not limit the number of dispatcher threads. The problem here is that the same dispatcher/thread at the client can pick a connection to the server from a pool, so even if the number of threads serving that connection at the server side is limited to one, subsequent calls might be handled concurrently. And if the number of connections in that pool would be limited to one (by setting the `maxGIOPConnectionPerServer` parameter to 1), this would prohibit callbacks from the server to the client. I have not found another a way to bind a dispatcher to a certain connection.

> Given the reordering issue, I am definitely NOT for this PR - in no circumstances. Limiting the number of threads per connection could very well make the performance situation worse, not better, as omniORB would not be allowed to spawn new threads to serve calls in parallel.

Switching to oneway write operations is not the only change introduced by this patch. I agree that the reordering issue needs to be revisited and perhaps we can revert this change for now, but the main contribution of this PR is a major reduction of the number of CORBA calls required to synchronize a TaskContextServer with its proxy, which is especially important to reduce setup time in high latency networks.

The main case where oneway writes are really necessary is if the publishing rate is higher than the reciprocal network round-trip time and the application cares more about reducing the latency between publisher and receiver than about not loosing messages or receiving them out of order. Without oneway writes, the dispatcher thread waits for a response for every single message sent and therefore can never be faster than the round-trip time and the intermediate buffer will fill up. The decision to prefer one over the other really depends on the use case and hence should be configurable.

doudou:
> Reducing the number of dispatchers in a process would not solve the problem.

We're not trying to solve the same "problem" here I think. Being able to partition the CORBA connections into different dispatchers would allow to isolate local connections from remote connections (provided that we allow omniorb to spawn enough threads), so that communication between same-host connections is not influenced by lag on high-latency connections to e.g. GUIs

meyerj:
Ah, now I got it. Same host is not equivalent to same process. In any case you did not want to share dispatchers between different components, but rather allow one component to be served by multiple dispatchers, right? That's what I understood.

I still do not see how this would solve the problem. If I do not care about the response, there is no need to spawn multiple dispatcher threads that do blocking writes or to handle connections to different peers differently. And if I do, multiple threads would allow some concurrency and solve the congestion problem reported by @planthaber in #122, but still cannot completely avoid congestion on a single connection without a way to send write operations oneway. Before #114 there was no way to get feedback about the result of a write operation anyway, and a failure during a remote write operation simply causes a disconnection and prints an error message.

Blocking/reliable vs. non-blocking/oneway/unreliable is a property of the connection, not of the dispatcher thread. For me, this is independent from the number of dispatcher threads or the allocation of dispatcher threads to connections, which would perhaps solve the problem of not blocking local connections in case of congestion in remote connections, but none of the two solutions solves both issues and it would be preferable to avoid congestion in the first place.

doudou:
> but still cannot completely avoid congestion on a single connection without a way to send write operations oneway

Even with oneway, if you can't configure one thread = one port-to-port connection explicitely, you don't avoid congestion on one port-to-port connection to affect the other port-to-port connections. Exposing the dispatcher configuration at the system design level at least allows you to trade off between this extreme (one dispatcher per connection) and the current situation (one dispatcher per component).

In the non-oneway case, one dispatcher serializes all the writes it is doing. One can therefore decide how many threads you allocate to omniorb and infer the type of behaviour to expect (for instance, allowing omniorb to create enough connections per server would have the effect to remove any cross-port-connections interactions).

With oneway, you could in principle have a single high-frequency output port on a high-latency link starve an arbitrarily big connection pool.

> Blocking/reliable vs. non-blocking/oneway/unreliable is a property of the connection

The problem to me with oneway is that it's not only "unreliable" (as: drops samples), but also unordered. Note that I'm not against having it optionally it's just IMO not usable in most cases.

meyerj:
I completely agree. Different problem, different solution.

I only understood the problem that different connections from the same component can influence each other in case of network congestion after I finished my last post, so the intention of having multiple dispatchers is clear now. But this was not the original goal of this PR and we should open another issue or reopen #122.

As a consequence, in order to not break existing applications and solve the reordering problem, I would again propose to extend the API and have two different write operations in the IDL (e.g. `write()` and `writeOneway()`) and let the RemoteChannelElement decide which one to call based on a new `reliability` flag in the connection policy. The `writeOneway()` method was already present in the [merge commit](https://github.com/orocos-toolchain/rtt/commit/5354f393d3c50c379948bc0476ceacba9ff55219) with the [updated-dataflow-semantics](https://github.com/orocos-toolchain/rtt/pull/114) branch in toolchain-2.9 to differentiate mandatory and non-mandatory connections and has been removed in https://github.com/orocos-toolchain/rtt/commit/3643898159eee5aa59f961dc90c5b54d240e09e4. Unreliable translates to "oneway" for the CORBA transport, and it would be up to the application developer to configure the CORBA layer and to understand the implications of using unreliable transports.

planthaber:
I agree to remove the oneway topic from this PR and reopen #122.

But in think @doudou is on the right track here.

As I understood, the Dispatcher holds the send queue for the port. When we now remove oneway completely and we have a dispatcher (and thread?) per connection we would also use two queues.
The problem I reported would also be solved: even if the remote connection is congested, messages are only dropped on the full queue for the congested connection, but not on the local one.

Is this what you meant Sylvain?

doudou:
> Is this what you meant Sylvain?

What I meant is that we should give the system designer the freedom to decide how to dispatch. In the extreme case where it creates one dispatcher per connection, yes you're basically right.

meyerj:
Re-merged into toolchain-2.9 in 171ed58a5f825552b95feb2e2acc0b5a69dddd59.

meyerj:
@planthaber @doudou Do you agree with merging this old patch along with a version bump to at least 2.9 into `master` and to follow-up on the remaining issues about partitioning the remote connections in https://github.com/orocos-toolchain/rtt/pull/227 and https://github.com/orocos-toolchain/rtt/pull/248?

This pull request is mainly about the more efficient communication of the TaskContext description and the oneway writes have been reverted due to reordering issues (#151) (unless RTT is compiled with `PLUGINS_CORBA_PORTS_WRITE_ONEWAY`).

planthaber:
I don't see an issue with the merge

planthaber:
I don't see an issue with the merge

planthaber:
I don't see an issue with the merge

doudou:
Sounds good to me too.

PR #127, opened 2016-01-05
Timer: fixed wake up time on 32 bit systems

pdima:
The existing comment is right:
We can't use infinite as the OS may internally use time_spec, which can not
represent as much in the future (until 2038) // XXX Year-2038 Bug

but used value of InfiniteNSecs/4 exceeds max in32 seconds.

Changed to use value to be converted within max int32 seconds.

### DIFF
--- a/rtt/os/Timer.cpp
+++ b/rtt/os/Timer.cpp
@@ -41,6 +41,7 @@
 #include "../Activity.hpp"
 #include "../Logger.hpp"
 #include "../os/fosi.h"
+#include <limits>
 
 namespace RTT {
     using namespace base;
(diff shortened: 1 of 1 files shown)


meyerj:
Nice catch. Did you experience any kind of problem on a 32 bit system?

Note: `long long wake_up_time` might be passed unmodified as an argument to `RTT::Semaphore::waitUntil()` and converted to
```
struct timespec {
    time_t  tv_sec;    // seconds
    long    tv_nsec;   // nanoseconds
}
```
in `ticks2timespec()`. The seconds part of of `wake_up_time` must therefore not exceed the range of a 32 bit integer and of type `time_t`, which happens to be of type `long` in Linux/gcc with a maximum value of `2^31 - 1` on 32 bit systems.

The `-1` part is IMHO not necessary here. We could also replace the whole line by
```
wake_up_time = 2147483647000000000LL;
```
for simplicity.

pdima:
I have seen existing value of (TimeService::InfiniteNSecs/4)-1 to point to time in the past when converted to 32bit so waiting loop started to use 100% of CPU once the last timer removed.

I agree -1 is not necessary, would you like me to update the commit?
Initially I had the same line as you proposed, but changed it to use std::numeric_limits<int32_t> to be more self documenting.

meyerj:
On Tue, Jan 12, 2016 at 10:44 PM, Dmytro Poplavskiy <
notifications@github.com> wrote:

> I have seen existing value of (TimeService::InfiniteNSecs/4)-1 to point to
> time in the past when converted to 32bit so waiting loop started to use
> 100% of CPU once the last timer removed.
>
> I agree -1 is not necessary, would you like me to update the commit?
>
Yes, please.


> Initially I had the same line as you proposed, but changed it to use
> std::numeric_limits to be more self documenting.
>
Fine for me. I like self-documenting code.

pdima:
Done

PR #128, opened 2016-01-05
Set thread name to match activity name on linux

pdima:
It simplifies debugging and allows to associate threads
with tasks/activities

### DIFF
--- a/rtt/os/gnulinux/fosi_internal.cpp
+++ b/rtt/os/gnulinux/fosi_internal.cpp
@@ -155,6 +155,16 @@ namespace RTT
 	    rv = pthread_create(&(task->thread), &(task->attr),
 	    		rtos_posix_thread_wrapper, xcookie);
 
+        // Set thread name to match task name, to help with debugging
+        {
+            // trim the name to fit 16 bytes restriction of pthread_setname_np
+            const int maxThreadNameSize = 16;
+            char threadName[maxThreadNameSize];
+            strncpy(threadName, task->name, maxThreadNameSize);
+            threadName[maxThreadNameSize-1] = 0;
+            pthread_setname_np(task->thread, threadName);
+        }
+
 	if ( cpu_affinity != (unsigned)~0 ) {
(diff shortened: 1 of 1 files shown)


meyerj:
looks good to me

jmachowinski:
Thread names are limited to 16 characters, I don't see, how the patch assures this

pdima:
Good point, pthread_setname_np returns ERANGE for larger names.
Would be better to truncate name to 16 chars, would you prefer such behavior?

meyerj:
On Tue, Jan 12, 2016 at 10:49 PM, Dmytro Poplavskiy <
notifications@github.com> wrote:

> Good point, pthread_setname_np returns ERANGE for larger names.
> Would be better to truncate name to 16 chars, would you prefer such
> behavior?
>
+1

jmachowinski:
Hm, I would propose a slight change, to use the last 16 characters.
If you use default orogen deployments, the are names orogen_module_task.
So the first 7 characters mostly don't give you any information.

meyerj:
I added commit bc05dceed2f123c35893bf0033d7257373f2f605 and pushed the updated [display_thread_names](https://github.com/orocos-toolchain/rtt/compare/display_thread_names) branch that limits the pthread name to the last characters of the Orocos task name as proposed by @jmachowinski and added a new test case to the task-test. Actually the maximum number of characters for the POSIX thread name is 15, because the terminating null byte has to be included.

It might be beneficial to change some standard thread names in RTT together with this patch, like:
* `"Activity"`
* `"TimerThreadInterface"` -> `"ThreadInterface"` (truncated)
* `"FileDescriptorActivity"` -> `"criptorActivity"` (truncated)
* ...

These examples actually show that neither the first nor the last 15 characters of a task name are a good choice as a POSIX thread name unless existing task names would be adapted, too.

Anyway, if you agree with this additional commit, I consider this PR as good to merge.

jmachowinski:
Looks good to me

PR #129, opened 2016-01-13
Skip task/timer updated if late for more than a few periods

pdima:
I'm not confident the proposed behavior is desired, I'm looking for comments/feedback.

I had the case with system time changed after rtt started, it caused all the periodic activities and timer to
run continuously to catch up all the missing updates since 1970.
I think the similar situation may happen when the system clock changed for other reasons like
daylight or time zone changes.

The better solution would be to use monotonic clocks, but as I understand it can't be used with sem_timedwait.

### DIFF
--- a/rtt/os/gnulinux/fosi_internal.cpp
+++ b/rtt/os/gnulinux/fosi_internal.cpp
@@ -254,9 +254,16 @@ namespace RTT
 
         if (task->wait_policy == ORO_WAIT_ABS)
         {
+          // in the case of overrun by more than 4 periods,
+          // skip all the updates before now, with the next update aligned to period
+          int maxDelayInPeriods = 4;
+          NANO_TIME period = task->period;
+          if (now - wake > maxDelayInPeriods*period) {
+            period = period * ((now - wake) / period);
+          }
           // program next period:
           // 1. convert period to timespec
-          TIME_SPEC ts = ticks2timespec( nano2ticks( task->period) );
+          TIME_SPEC ts = ticks2timespec( nano2ticks(period) );
           // 2. Add ts to periodMark (danger: tn guards for overflows!)
           NANO_TIME tn = (task->periodMark.tv_nsec + ts.tv_nsec);
           task->periodMark.tv_nsec = tn % 1000000000LL;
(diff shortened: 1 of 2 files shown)


meyerj:
Just for the record: There are plans to rewrite the Thread and Timer classes to not depend on real-time clock (system time) anymore, but to use condition variables and monotonic clock sources instead whenever applicable. This would also solve the problem that periodic tasks cannot be stopped before the current waiting period expired.

But as long as we do not have that patch, this PR is a step in the right direction. Nevertheless it should be considered as a workaround. Thanks for the effort!

rtt/os/gnulinux/fosi_internal.cpp
```diff
@@ -254,9 +254,16 @@ namespace RTT
 
         if (task->wait_policy == ORO_WAIT_ABS)
         {
+          // in the case of overrun by more than 4 periods,
+          // skip all the updates before now, with the next update aligned to period
+          int maxDelayInPeriods = 4;
+          NANO_TIME period = task->period;
+          if (now - wake > maxDelayInPeriods*period) {
+            period = period * ((now - wake) / period);
```
meyerj: Actually this should have been
```
period += period * ((now - wake) / period);
```
because we always want to add at least one period ([line 260](https://github.com/orocos-toolchain/rtt/pull/129/files#diff-2f91e9fcda1ad4baa9b66f9cc15b9c88R260)) plus eventually the additional ones, like in [Timer.cpp:108](https://github.com/orocos-toolchain/rtt/pull/129/files#diff-8741c88e5586ddedd496eba04289892eR108).

meyerj:
I am thinking about reverting this patch for toolchain-2.9 in favor of https://github.com/orocos-toolchain/rtt/pull/138. #138 only provides a solution for the `Thread` class, but not for `Timer`.

Also it should be noted that both patches, #129 and #138, are not effective anymore for periodic components in connection with #91. So actually we would need a clean solution for all these cases of interruptible, perodic waiting by adding an OS abstraction of condition variables which support waiting based on a steady clock source. For gnulinux/pthread that would be using [pthread_condattr_setclock](http://pubs.opengroup.org/onlinepubs/9699919799/) with `CLOCK_MONOTONIC`. And get rid of the semaphore in `os::Timer` by using a condition variable instead.

PR #136, opened 2016-02-04
add getName method

ahoarau:
Just a getName method.
Don't know what happens with my editor and that last line.

### DIFF
--- a/rtt/TaskContext.cpp
+++ b/rtt/TaskContext.cpp
@@ -99,6 +99,7 @@ namespace RTT
         // from Service
         provides()->doc("The interface of this TaskContext.");
 
+        this->addOperation("getName", &TaskContext::getName, this, ClientThread).doc("The name of this TaskContext.");
         this->addOperation("configure", &TaskContext::configure, this, ClientThread).doc("Configure this TaskContext (= configureHook() ).");
         this->addOperation("isConfigured", &TaskContext::isConfigured, this, ClientThread).doc("Is this TaskContext configured ?");
         this->addOperation("start", &TaskContext::start, this, ClientThread).doc("Start this TaskContext (= startHook() + updateHook() ).");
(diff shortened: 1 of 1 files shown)


meyerj:
@ahoarau: Could you shortly motivate why this operation is needed?

@smits: Can this have any side effects in connection with the Corba aliases (#66)? The aliased remote proxy of a TaskContext would still return the original name from the `getName()` call.

ahoarau:
@meyerj I'm facing difficulties with the [services](https://github.com/orocos-toolchain/ocl/issues/40).
The only two things (I found) that can load the [owner's ptr to the service](https://github.com/orocos-toolchain/rtt/blob/3720afb0726688a0bb46d68e7f96fbb72ef10f5d/rtt/plugin/ServicePlugin.hpp#L143:L144) are ```loadService("Deployer","rosdeployment")``` or ```.provides rosdeployment``` but the latter one does not work if called from ```runScript(/the/script/that/contains/provide/rosdeployment)```.

So I can only use ```loadService("Deployer","rosdeployment")```. 
I wish I could do ```this.loadService("rosdeployment")``` with [this function](https://github.com/orocos-toolchain/rtt/blob/d2bc6f40983bef71f92dc6d45a9fc4d2b75ba2a3/rtt/TaskContext.cpp#L234) but it gets [overridden](https://github.com/orocos-toolchain/ocl/blob/a4bea41d6226e9024b3867b898f1ebbd30dc4302/deployment/DeploymentComponent.cpp#L132:L134).

meyerj:
Can this PR be closed with the solution from https://github.com/orocos-toolchain/ocl/pull/42 or is there another use case where a `getName()` operation might be useful?

ahoarau:
Perfectly fine with the "this" solution, thanks ! closing PR.

PR #137, opened 2016-02-09
State machine bug fixes

meyerj:
This PR contains a subset of the patches in #135, that only fix bugs and do not change the execution semantics of Orocos state machines. We consider to only include those in the 2.9 release but not yet the new execution logic.

167076a68ced2fad11a0998ef4df76ade1b31200 solves an issue where the entry program of a state can be aborted by an event in some cases. dc1b39b86334ac0a6a897590dce689b8195e7ccd removes a duplicate call to `enableEvents()`, which would trigger an assert if RTT is compiled in debug mode. The other two are minor patches for the unit tests.

### DIFF
--- a/rtt/scripting/StateMachine.cpp
+++ b/rtt/scripting/StateMachine.cpp
@@ -327,8 +327,10 @@ namespace RTT {
             this->executePending();
             break;
         case Status::running:
-            if ( this->executePending() == false)
+            if ( this->executePending() == false) {
+                TRACE("Yielding...");
                 break;
+            }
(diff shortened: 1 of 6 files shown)


meyerj:
Merged to [toolchain-2.9](https://github.com/orocos-toolchain/rtt/tree/toolchain-2.9) branch in 8e4cfc68194b788b3c7ff7dd6b05e9d829ae1ec6 and 3acb4671e2b6c56fb55c95576f208928d010afe3.

PR #139, opened 2016-02-09
pluginloader: fixed segfault if a service plugin refuses to create a service instance to be loaded into the global service

meyerj:
Originally reported in orocos-toolchain/ocl#40.

To reproduce:

```
$ deployer
import("rtt_rosdeployment")
require("rosdeployment")
```

The problem is that `createService()` returns a null pointer in [rtt_rosdeployment_service.cpp](https://github.com/orocos/rtt_ros_integration/blob/indigo-devel/rtt_rosdeployment/src/rtt_rosdeployment_service.cpp) and the PluginLoader did not catch this case before forwarding it to `Service::addService()`.


meyerj:
Merged into toolchain-2.9: 6b0963158f5d1a4d92b23a8b1741ce106fa4b743

PR #142, opened 2016-03-18
Implement CORBA disconnect(port).

eberghoefer:
Disconnect from certain port.

Implementing the previous missing function to disconnect RTT::corba::RemoteOutputPort and RTT::corba::RemoteInputPorts from a certain given remote port.
(instead of disconnect all).

### DIFF
--- a/rtt/transports/corba/RemotePorts.cpp
+++ b/rtt/transports/corba/RemotePorts.cpp
@@ -47,7 +47,6 @@
 #include "../../internal/ConnID.hpp"
 #include "../../rtt-detail-fwd.hpp"
 
-
 using namespace std;
 using namespace RTT::detail;
 
@@ -78,13 +77,8 @@ void RemotePort<BaseClass>::disconnect()
 {
     dataflow->disconnectPort(this->getName().c_str());
 }
-template<typename BaseClass>
-bool RemotePort<BaseClass>::disconnect(PortInterface* port)
-{
-    Logger::In in("RemotePort::disconnect(PortInterface& port)");
-    log(Error) << "Disconnecting a single port not yet supported." <<endlog();
-    return false;
-}
+
+
 template<typename BaseClass>
 PortableServer::POA_ptr RemotePort<BaseClass>::_default_POA()
 { return PortableServer::POA::_duplicate(mpoa); }
@@ -194,6 +188,12 @@ RTT::base::ChannelElementBase::shared_ptr RemoteInputPort::buildRemoteChannelOut
     return corba_ceb;
 }
 
+bool RemoteInputPort::disconnect(PortInterface* port)
+{
+    //just calling the implementation from the other side. (implemented already for RemoteOutputPort).
+    return port->disconnect(this);
+}
+
 RTT::base::PortInterface* RemoteInputPort::clone() const
 { return type_info->inputPort(getName()); }
 
@@ -235,6 +235,26 @@ void RemoteOutputPort::keepLastWrittenValue(bool new_flag)
 DataSourceBase::shared_ptr RemoteOutputPort::getDataSource() const
 {
     return DataSourceBase::shared_ptr();
+}
+
+
+bool RemoteOutputPort::disconnect(PortInterface* port)
+{
+    RemoteInputPort *portI = dynamic_cast<RemoteInputPort *>(port);
+
+    //if not a remote port, we can not handle at the moment!
+    if(portI == NULL){
+        Logger::In in("RemoteOutputPort::disconnect(PortInterface& port)");
+        log(Error) << "Port: " << port->getName() << " could not be disconnected from: " << this->getName()
+                   << " because it could not be casted to a RemoteInputPort type!" << nlog()
+                   << "Only disconnect of two remote ports supported by corba layer, yet!" << endlog();
+        return false;
+
+    }
+
+    //if Remote Input Port:
+    return dataflow->removeConnection(this->getName().c_str(), portI->getDataFlowInterface(),
+                                      portI->getName().c_str());
 }
 
 bool RemoteOutputPort::createConnection( RTT::base::InputPortInterface& sink, RTT::ConnPolicy const& policy )
--- a/tests/corba_test.cpp
+++ b/tests/corba_test.cpp
@@ -598,6 +598,95 @@ BOOST_AUTO_TEST_CASE( testPortProxying )
     mo2->disconnect();
 }
 
+BOOST_AUTO_TEST_CASE( testRemotePortDisconnect )
+{
+    ts  = corba::TaskContextServer::Create( tc, false ); //no-naming
+    tp  = corba::TaskContextProxy::Create( ts->server(), true );
+    ts2  = corba::TaskContextServer::Create( t2, false ); //no-naming
+    tp2  = corba::TaskContextProxy::Create( ts2->server(), true );
+
+    // Create a default CORBA policy specification
+    RTT::ConnPolicy policy = ConnPolicy::data();
+    policy.init = false;
+    policy.transport = ORO_CORBA_PROTOCOL_ID; // force creation of non-local connections
+
+    base::PortInterface* untyped_port;
+
+    //mi1
+    untyped_port = tp->ports()->getPort("mi");
+    BOOST_CHECK(untyped_port);
+    base::InputPortInterface* read_port1 = dynamic_cast<base::InputPortInterface*>(tp->ports()->getPort("mi"));
+    BOOST_CHECK(read_port1);
+
+    //mi2
+    untyped_port = tp2->ports()->getPort("mi");
+    BOOST_CHECK(untyped_port);
+    base::InputPortInterface* read_port2 = dynamic_cast<base::InputPortInterface*>(tp2->ports()->getPort("mi"));
+    BOOST_CHECK(read_port2);
+
+    //mo2
+    untyped_port = tp2->ports()->getPort("mo");
+    BOOST_CHECK(untyped_port);
+    base::OutputPortInterface* write_port2 = dynamic_cast<base::OutputPortInterface*>(tp2->ports()->getPort("mo"));
+    BOOST_CHECK(write_port2);
+
+    //mo1
+    untyped_port = tp->ports()->getPort("mo");
+    BOOST_CHECK(untyped_port);
+    base::OutputPortInterface* write_port1 = dynamic_cast<base::OutputPortInterface*>(tp->ports()->getPort("mo"));
+    BOOST_CHECK(write_port1);
+
+    // Just make sure 'read_port' and 'write_port' are actually proxies and not
+    // the real thing
+    BOOST_CHECK(dynamic_cast<corba::RemoteInputPort*>(read_port1));
+    BOOST_CHECK(dynamic_cast<corba::RemoteInputPort*>(read_port2));
+    BOOST_CHECK(dynamic_cast<corba::RemoteOutputPort*>(write_port1));
+    BOOST_CHECK(dynamic_cast<corba::RemoteOutputPort*>(write_port2));
+
+    //create remote connections:
+    write_port2->connectTo(read_port1, policy);
+    BOOST_CHECK(read_port1->connected());
+    BOOST_CHECK(write_port2->connected());
+
+    //second connection to this input port
+    write_port1->connectTo(read_port1, policy);
+    BOOST_CHECK(write_port1->connected());
+
+    //disconnect single port connection (both remote), same tcs object.
+    write_port1->disconnect(read_port1);
+    BOOST_CHECK(read_port1->connected());
+    BOOST_CHECK(!write_port1->connected());
+
+    //disconnect single port connection, both remote, different tcs.
+    write_port2->disconnect(read_port1);
+    BOOST_CHECK(!read_port1->connected());
+
+    //check disconnecting call on reader port. (build connection again beforehand).
+    write_port2->connectTo(read_port1, policy);
+    BOOST_CHECK(read_port1->connected());
+    BOOST_CHECK(write_port2->connected());
+    BOOST_CHECK(read_port1->disconnect(write_port2));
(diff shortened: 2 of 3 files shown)


rtt/transports/corba/RemotePorts.cpp
```diff
@@ -194,6 +189,20 @@ RTT::base::ChannelElementBase::shared_ptr RemoteInputPort::buildRemoteChannelOut
...
+    //just check the type and not relay on
+    RemoteOutputPort *oPort = dynamic_cast<RemoteOutputPort* >(port);
+    if(oPort == NULL){
+        Logger::In in("RemoteInputPort::disconnect(PortInterface& port)");
+        log(Error) << "Port: " << this->getName() << " could not be disconnected from: " << oPort->getName()
+                << " because it is not of type: RemoteOutputPort" << endlog();
+        return false;
+    }
+
+    return dataflow->removeConnection(oPort->getName().c_str(), this->getDataFlowInterface(), this->getName().c_str());
```
doudou: This looks weird. Doesn't `this->getDataFlowInterface() == dataflow`. In which case, don't you basically require oPort and this to be from the same data flow interface ?

Moreover, you only support disconnection between two remote ports. Wouldn't there be a possibility to support disconnection from a remote port to a local one ?

eberghoefer: @doudou, true you are right, could be simplified by just using "dataflow" here.

Actually I was simply not sure if the port has to be Remote, It was a guess based on the other implementations. So would it allow to be a local port as well by using.

Also just wondering if 
`return port.disconnect(this)`
could do the job here for all cases?

jmachowinski: If you implement RemoteInputPort like this :
return dataflow->removeConnection(port->getName().c_str(), dataflow, getName().c_str());

You should be able to implement the RemoteOutpuPort as 
return port.disconnect(this)

If I'm not missing something.

doudou: > @doudou, true you are right, could be simplified by just using "dataflow" here.

That is definitely NOT what I implied here.

My point is: if the dataflow interface on which you call removeConnection and the one you give to the method are the same, I believe that it implies that both ports are managed by the *same* dataflow interface. Meaning that they come from the same component. What am I missing ?

jmachowinski: You need to pass the reader interface to the function :
`bool removeConnection( const char* writer_port,
                                               CDataFlowInterface_ptr reader_interface,
                                               const char* reader_port) ACE_THROW_SPEC ((
                                                      CORBA::SystemException
                                                      ,::RTT::corba::CNoSuchPortException
                                                    ));
`

doudou: reader_interface is indeed the interface of reader_port. And unless I am mistaken, the receiver is the dataflow interface of writer_port. In this patch, both dataflow interface objects are the same, which implies that the writer and reader are from the same dataflow interface.

In any case, there should be unit tests added for this.

eberghoefer: @doudou, yes I got that. But in the patch I implemented both, disconnecto for RemoteOutput and RemoteInputPort. So since the data flow remove connection method uses a pointer to the data flow oject of "the other" port, to me it looks more natural to do that from the "view" of the output port. Anyway wouldn't it be better to have the code only once instead of twice? 
In the RemoteOutputPort than it is also more natural, that the dataflow object on which the function is called is the local one and the pointer given as the remote dataflow object is the one from the RemoteInputPort. So if just calling it in the RemoteInputPort by calling the function of the OutputPort to disconnect from it, and therefore, prevent to have very similar code at two different positions doesn't look too bad for me. Also then on this side it would directly call the right implementation if the given port is a local one.

However, then I guess it moves the question of how to deal with a "local" InputPort given as parameter to the `RemoteOutputPort.disconnect()` function. Is that even a case that can occur? If so, I do not really see how to handle it other than calling again `disconnect(this)` on the port if it is local hoping it can deal with remote ports than, but could call loops here then, which would be bad. But since the FlowInteface needed for the removeConnection function is not provided by an InputPort and to its connectionManager it can only provide a read only (const) version. 
So what is the right way to go for that?

doudou: > more natural to do that from the "view" of the output port

You really are not explaining anything to me. What's natural is currently irrelevant, I don't even understand why it should *work*.

Let's settle this one easy way: write unit tests, we'll see what works and what does not.

eberghoefer:
I extended the pull request with an additional unit-test in the corba-test application.
For my use-case only the call of the disconnect function on RemotePorts giving also another RemotePort as parameter is of interest. The according unit-tests are successful.

Since @doudou requested also having a possibility of giving a local port as parameter, I added two tests for that too.
Unfortunately I couldn't figure out how to make that case work (see my next comment with code reference), so suggestions on that one would be very welcome.

rtt/transports/corba/RemotePorts.cpp
```diff
@@ -252,12 +242,23 @@ bool RemoteOutputPort::disconnect(PortInterface* port)
...
+        if(portIn == NULL){
+            Logger::In in("RemoteOutputPort::disconnect(PortInterface& port)");
+            log(Error) << "Port: " << port->getName() << " could not be disconnected from: " << this->getName()
+                    << "because it could not be casted to an InputPort type!" << endlog();
+            return false;
+        }
+
+        if(portIn->disconnect(this) or cmanager.disconnect(portIn)){
+            return true;
+        }
```
eberghoefer: So here is the problem, neither the local port
`portIn->disconnect(this)`
seem to be capable of disconnecting itself from a remote port nor
`cmanager.disconnect(portIn))`
the connection manager of the remote port can disconnect from the local port.
Since the flow interface disconnect function requires the "other" port to have a CDataFlowInterface as well, it can not be applied on the local port as also. So at least to me there was no obvious way how to realise the disconnect then.
Any suggestions?

rtt/transports/corba/RemotePorts.cpp
```diff
@@ -237,6 +236,32 @@ DataSourceBase::shared_ptr RemoteOutputPort::getDataSource() const
...
+        RTT::base::InputPortInterface *portIn = dynamic_cast<RTT::base::InputPortInterface *>(port);
+        //if still NULL give up:
+        if(portIn == NULL){
+            Logger::In in("RemoteOutputPort::disconnect(PortInterface& port)");
+            log(Error) << "Port: " << port->getName() << " could not be disconnected from: " << this->getName()
+                    << "because it could not be casted to an InputPort type!" << endlog();
+            return false;
+        }
+
+        if(portIn->disconnect(this) or cmanager.disconnect(portIn)){
```
eberghoefer: So here is the problem, neither the local port
`portIn->disconnect(this)`
seem to be capable of disconnecting itself from a remote port nor
`cmanager.disconnect(portIn))`
the connection manager of the remote port can disconnect from the local port.
Since the flow interface disconnect function requires the "other" port to have a CDataFlowInterface as well, it can not be applied on the local port as also. So at least to me there was no obvious way how to realise the disconnect then.
Any suggestions?

doudou:
> Any suggestions?

Well, since it seems that it would not be possible to get local port suport easily, I think it's fair to reduce the functionality at disconnection of remote ports. But then you should test for that case and emit an exception (or return false ? I don't know what would be closer to RTT's general behaviour there)

eberghoefer:
Yes I agree, actually I think that the use case to disconnect local from remote ports or vice versa will still be a rare corner case.
Since at least in the RTT::corba stuff it seams to be that all methods that have a return bool signalling errors by returning false and writing an error message to the log, I followed that here too and write a message to the log and return false, if the "other" port given to one of the RemotePort Disconnect(Port) functions is not a RemotePort too. 
Also to make clear that this is the case and make sure that indeed false is returned, I leave the unitTest in that checks local to remote variants, but expecting a false as return now.

eberghoefer:
So if no further objections, could someone please merge the request.

doudou:
@jbohren @meyerj if you could have a look ...

tests/corba_test.cpp
```diff
@@ -598,6 +598,90 @@ BOOST_AUTO_TEST_CASE( testPortProxying )
...
+    //check disconnecting call on reader port. (build connection again beforehand).
+    write_port2->createConnection(*read_port1);
+    BOOST_CHECK(read_port1->connected());
+    BOOST_CHECK(write_port2->connected());
+    BOOST_CHECK(read_port1->disconnect(write_port2));
+    BOOST_CHECK(!read_port1->connected());
+    BOOST_CHECK(!write_port2->connected());
+
+    //check connect and disconnect certain port, remote output to local input
+    //should give falls cause not supported yet!
```
meyerj: typo: false

tests/corba_test.cpp
```diff
@@ -598,6 +598,90 @@ BOOST_AUTO_TEST_CASE( testPortProxying )
...
+    write_port2->connectTo(mi1);
+    BOOST_CHECK(mi1->connected());
+    BOOST_CHECK(write_port2->connected());
+    BOOST_CHECK(!write_port2->disconnect(mi1));
+    mi1->disconnect();
+
+    //check disconnect remote input port from local output port.
+    //should give falls cause not supported yet!
```
meyerj: typo: false

rtt/transports/corba/RemotePorts.cpp
```diff
@@ -237,6 +245,25 @@ DataSourceBase::shared_ptr RemoteOutputPort::getDataSource() const
...
 
+
+bool RemoteOutputPort::disconnect(PortInterface* port)
+{
+    RemoteInputPort *portI = dynamic_cast<RemoteInputPort *>(port);
+
+    //if not a remote port, we can not handle at the moment!
+    if(portI == NULL){
+        Logger::In in("RemoteOutputPort::disconnect(PortInterface& port)");
+        log(Error) << "Port: " << port->getName() << " could not be disconnected from: " << this->getName()
```
meyerj: missing space or `nlog()` after port name:
```
1.230 [ ERROR  ][RemoteOutputPort::disconnect(PortInterface& port)] Port: mi could not be disconnected from: mobecause it could not be casted to an RemoteInputPort type!
Only disconnect of two remote ports supported by corba layer, yet!

```

RTT Logger has a `nlog()` method (see documentation [here](http://orocos-toolchain.github.io/devel/documentation/rtt/master/api/html/classRTT_1_1Logger.html#ac6fabc2ad9f34905a576dff32952e173)), which seems to be preferred over `std::endl`.

rtt/transports/corba/RemotePorts.cpp
```diff
@@ -237,6 +245,25 @@ DataSourceBase::shared_ptr RemoteOutputPort::getDataSource() const
...
+                   << "because it could not be casted to an RemoteInputPort type!" << std::endl
+                   << "Only disconnect of two remote ports supported by corba layer, yet!" << std::endl << endlog();
```
meyerj: Please remove `std::endl` just before `endlog()` to not generate an empty line.

rtt/transports/corba/RemotePorts.cpp
```diff
@@ -194,6 +188,20 @@ RTT::base::ChannelElementBase::shared_ptr RemoteInputPort::buildRemoteChannelOut
...
 
+bool RemoteInputPort::disconnect(PortInterface* port)
+{
+    //only try if other port is remote too:
+    RemoteOutputPort *oPort = dynamic_cast<RemoteOutputPort *>(port);
+    if(oPort == NULL){
+        Logger::In in("RemoteInputPort::disconnect(PortInterface *port)");
+        log(Error) << "Port: " << port->getName() << " could not be disconnected from: " << this->getName()
+                   << "because it could not be casted to an RemoteOutputPort type!" << std::endl
+                   << "Only disconnect of two remote ports supported by corba layer, yet!" << std::endl << endlog();
```
meyerj: See below: new-line/space formatting

meyerj:
I did a quick pass and added some nit comments to the PR.

The new test `testRemotePortDisconnect` was actually not creating true remote connections because `DataFlowI::createConnection()` bypasses the creation of a remote connection and creates a local connection instead if the reader port is local (from the perspective of the output port). This could have an effect on the disconnection, too. I added a commit in my fork that explicitly sets the transport_id in the connection policy in order to enforce the CORBA transport, same as other tests: https://github.com/meyerj/rtt/commit/ed9348e77a3c01216c19c1fbee4ee470fc7710f4

The case where one of the two ports is local could simply be covered by forwarding to the local port's `disconnect(port)` implementation, which keep track of the remote interfaces in the `RemoteConnId` stored inside the connection manager. I have not verified this.

eberghoefer:
@meyerj Thank you for your review, I have changed your requests and picked the corrected unitTest from your fork (thanks for that one too ;) ).

>         The case where one of the two ports is local could simply be covered by forwarding to the local 
>         port's disconnect(port) implementation, ...

This in principle was what I tried out first (can be found in the history of this discussion). 
Anyway, I tried it again with your adapted unit test, and it indeed changed something.
Now the `port.disconnect(this)` for the RemoteInputPort works also for the case where it is connected to a local output port. I have changed the unit-test again to expect this now and removed the log message for that reason!
But still, for the case RemoteOutputPort tries to disconnect from a local input port by trying to leave that up to the local port side `port.disconnect(this))` still does not work.
If you replace Line 251 (from the current version) of the file RemotePorts.cpp
with `return port->disconnect(this);` instead of just `return false;` you'll see, that it still returns "false" as the test expects. And adding a test here, that checks for the port being disconnected would fail.

So if you have an idea how to make this also work, let me know.

eberghoefer:
So since at least to have the functionality of disconnecting remote ports not requiring to stay on the branch, could anybody be so kind to merge the request? @meyerj @doudou ?
If not somebody has further requests to be changed.

jmachowinski:
Could you smash all commits into one ?

eberghoefer:
sure, here it goes...

PR #143, opened 2016-03-31
Fixed build when sequential activity is used by default

pdima:
Build fails when configured to use sequential activity by default, trivial fix


meyerj:
Also cherry-picked to master in https://github.com/orocos-toolchain/rtt/commit/08f0769b29eef12843697fe66d84f83196b7edd5.

PR #144, opened 2016-04-05
useorocos: added optional macro argument EXPORT and export cmake target names for all target types

meyerj:
CMake target importing and exporting [1] is an alternative way to find and link to targets that have been generated and installed from another package in case users do not want to rely on the `orocos_generate_package()` and `orocos_use_package()` macros provided by UseOROCOS-RTT.cmake. This PR supports this workflow by adding an EXPORT argument to the macros that create Orocos targets and by exporting all targets under the export name `${PROJECT_NAME}-${OROCOS_TARGET}` by default.

[1] https://cmake.org/Wiki/CMake/Tutorials/Exporting_and_Importing_Targets#Exporting_from_an_Installation_Tree

### DIFF
--- a/UseOROCOS-RTT.cmake
+++ b/UseOROCOS-RTT.cmake
@@ -236,7 +236,7 @@ if(OROCOS-RTT_FOUND AND NOT USE_OROCOS_RTT)
   #
   macro( orocos_component COMPONENT_NAME )
     ORO_PARSE_ARGUMENTS(ADD_COMPONENT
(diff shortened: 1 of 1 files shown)


meyerj:
Merged to [toolchain-2.9](https://github.com/orocos-toolchain/rtt/tree/toolchain-2.9) branch in bbacf2f9c6cf9b33335f9e79e580b2fb52d23738.

PR #145, opened 2016-04-05
useorocos: check the real path of the current and project source dir for automatic Orocos package name derivation

meyerj:
This is required if at least one component of the path to the source directory is a symlink, because `PROJECT_SOURCE_DIR` points to the symlinked location and `CMAKE_CURRENT_SOURCE_DIR` contains the real path of the project's source directory.

TODO: Abort with an error if CMAKE_CURRENT_SOURCE_DIR is not a subfolder of `PROJECT_SOURCE_DIR` and no package name was given explicitly.


meyerj:
Merged to [toolchain-2.9](https://github.com/orocos-toolchain/rtt/tree/toolchain-2.9) branch in 86dfa2f7b13447515f2328c544ba80a87098d8c6.

PR #146, opened 2016-04-05
Fixed typekit for bool sequences

meyerj:
This PR adds the missing implementation of specialized method `RTT::types::get_container_item_copy(const std::vector<bool> & cont, int index)` required by `SequenceTypeInfo<T>` implementation for type `std::vector<bool>`.

The default RTT typekit does not contain type `std::vector<bool>` and therefore does not trigger this error.

### DIFF
--- a/rtt/types/SequenceTypeInfoBase.cpp
+++ b/rtt/types/SequenceTypeInfoBase.cpp
@@ -48,5 +48,12 @@ namespace RTT
                 return false;
             return cont[index];
         }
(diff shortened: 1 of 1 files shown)


meyerj:
Merged to [toolchain-2.9](https://github.com/orocos-toolchain/rtt/tree/toolchain-2.9) branch in c55d8e150e115e3f933171b50e85500a4d75e3eb.

PR #148, opened 2016-04-11
Fixed potential double deallocation and pool corruption for circular BufferLockFree

meyerj:
The missing check of the return value of `bufs.dequeue()` let the buffers_test fail from time to time under high CPU load because of memory pool corruption due to double deallocation.

Merged to toolchain-2.9 in https://github.com/orocos-toolchain/rtt/commit/dcc7df5dd7685b167f337e29a8fb99000c757f84.

### DIFF
--- a/rtt/base/BufferLockFree.hpp
+++ b/rtt/base/BufferLockFree.hpp
@@ -170,8 +170,15 @@ namespace RTT
                     // pop & deallocate until we have free space.
                     Item* itmp = 0;
                     do {
-                        bufs.dequeue( itmp );
-                        mpool.deallocate( itmp );
+                        if ( bufs.dequeue( itmp ) ) {
+                            mpool.deallocate( itmp );
+                        } else {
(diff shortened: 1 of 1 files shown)


rtt/base/BufferLockFree.hpp
```diff
@@ -185,9 +185,15 @@ namespace RTT
                     // pop & deallocate until we have free space.
                     Item* itmp = 0;
                     do {
-                        bufs.dequeue( itmp );
-                        mpool.deallocate( itmp );
-                        droppedSamples.inc();
```
meyerj: For circular buffers samples are never dropped. We just keep on trying...

meyerj: Reverted `droppedSamples.inc()` in https://github.com/orocos-toolchain/rtt/commit/7fde6cbe677dbc46e58e8cdc2b9953da19fb0c2e: The new sample is not dropped, but the one removed from the circular buffer before. So incrementing the counter here is correct.

PR #151, opened 2016-05-12
corba: revert oneway writes and fixed connection cleanup and TAO builds

meyerj:
This PR adds some patches that fix the behavior of the CORBA transport in the toolchain-2.9 branch, which is a merge of various features, especially the efficiency patches (#123) and the updated dataflow semantics (#114).

The first commit 63218b0 fixes an issue that broke the proper cleanup after one end of a CORBA connection terminates or crashes. Now the `RemoteChannelElement::write()` call would return the WriteStatus `NotConnected` and thus indirectly, via the dispatcher, cause the writing side to remove the connection from the port. This is basically the behavior of RTT versions before 2.9. It is up to the user to eventually reestablish broken connections if the port's write call returns `NotConnected` in case this is required.

The second commit f69a0b7 reintroduces the standard blocking two-ways CORBA calls unless RTT has been compiled with the cmake option `PLUGINS_CORBA_PORTS_WRITE_ONEWAY` enabled explicitly. See also discussion in #122 and #123 for further information on this issue. It is up to a future update to make this a run-time option, e.g. by adding a boolean field `reliable` to the ConnPolicy struct.

The last commit bc6503b fixes some small issues if RTT is compiled with TAO as underlying CORBA implementation and corrects some whitespace inconsistencies. It has been tested with TAO 2.3.1 and all the unit tests run successfully now.

### DIFF
--- a/config/FindTAO.cmake
+++ b/config/FindTAO.cmake
@@ -193,7 +193,7 @@ MACRO(ORO_ADD_CORBA_SERVERS _sources _headers)
       GET_FILENAME_COMPONENT(_filedir ${_tmp_FILE} PATH)
 
       SET(_server  ${CMAKE_CURRENT_BINARY_DIR}/${_basename}S.cpp)
-      SET(_serverh ${CMAKE_CURRENT_BINARY_DIR}/${_basename}S.h ${CMAKE_CURRENT_BINARY_DIR}/${_basename}S.inl)
+      SET(_serverh ${CMAKE_CURRENT_BINARY_DIR}/${_basename}S.h)
 
       set(DEFINE_TAO "-DCORBA_IS_TAO")
       # From TAO 1.5 onwards, the _T files are no longer generated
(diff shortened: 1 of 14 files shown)


config/FindTAO.cmake
```diff
@@ -193,7 +193,7 @@ MACRO(ORO_ADD_CORBA_SERVERS _sources _headers)
       GET_FILENAME_COMPONENT(_filedir ${_tmp_FILE} PATH)
 
       SET(_server  ${CMAKE_CURRENT_BINARY_DIR}/${_basename}S.cpp)
-      SET(_serverh ${CMAKE_CURRENT_BINARY_DIR}/${_basename}S.h ${CMAKE_CURRENT_BINARY_DIR}/${_basename}S.inl)
+      SET(_serverh ${CMAKE_CURRENT_BINARY_DIR}/${_basename}S.h)
```
meyerj: This change was required on my machine with TAO 2.3.1 (Ubuntu 14.04 x64 with default compilers) to fix cmake errors when it tries to install non-existent files. I am not sure if older versions of TAO generate the server inline skeleton files and if they really need to be installed. At the end, this is the only effect of adding them to the `_serverh` variable here.

PR #153, opened 2016-05-13
Moved instantiation of DataSource template types for bool and std::string from the typekit to the core library

meyerj:
This is for consistency reasons and to avoid duplicate code generation.
The RTT core already uses various bool and std::string data sources internally and hence already has
instantiations of these template classes. This patch adds extern template declarations to the DataSource.hpp and DataSources.hpp headers and removes them from the typekit headers.

Also the instantiation of the rt_string and rt_ostringstream classes (not of the respective RTT template
instantiations) has been moved from the typekit to the core library for similar reasons.

This patch removes the GCC compiler warning `type attributes ignored after type is already defined [-Wattributes]` while compiling the typekit library (from Types.inc) because of inconsistent visibility attributes.

### DIFF
--- a/rtt/internal/DataSource.hpp
+++ b/rtt/internal/DataSource.hpp
@@ -236,5 +236,15 @@ namespace RTT
 #ifndef ORO_CORELIB_DATASOURCES_HPP
 #include "DataSource.inl"
 #endif
+
+/*
+ * Extern template declarations for core data source types
+ * (instantiated in DataSources.cpp)
+ */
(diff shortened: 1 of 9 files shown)


meyerj:
Merged to [toolchain-2.9](https://github.com/orocos-toolchain/rtt/tree/toolchain-2.9) branch in 6ccb412fe4e6a51751b35f497fe6abbebd2d0930.

meyerj:
This patch was already merged in the duplicate #234.

PR #154, opened 2016-06-02
copied gcc version matching part from master

planthaber:
stable does not even try to compile in newer OSes on the rock buildvserver, because gcc5 was detected as < gcc3 by cmake


meyerj:
Could we remove the gcc version check completely? Does anyone still uses a gcc version older than 3.3, where the visibility support has been added?

If the compiler check in CMakeLists.txt would be removed, the value of the `RTT_GCC_HASVISIBILITY` cmake variable and preprocessor macro must be considered as true and removed from the generated header files.

meyerj:
Merged into master in 6530c73e1db208b5bddbf3fe27c702f433e409e9.

PR #157, opened 2016-06-22
useorocos: fix RPATH issues when linking to ocl or 3rd-party libraries

meyerj:
This PR restores the INSTALL_RPATH behavior of RTT version 2.7 and before, where all directories in the linker's search path have been added to the INSTALL_RPATH of a target. #51 and #79 introduced DESTDIR support (installation in stage dir), where the locations of the libraries at link time is not necessarily the same as the installed location.

In all cases where OROCOS targets link to other libraries which are not installed in a directory within the default search path and without the `orocos_use_package()` macro, these patches broke dynamic linking without special care in the user's cmake code.

Example:
```cmake
find_package(OROCOS-OCL REQUIRED COMPONENTS ocl-logging ocl-timer)

orocos_component(foo foo.cpp)
target_link_libraries(foo ${OROCOS-OCL_OCL-LOGGING_LIBRARIES} ${OROCOS-OCL_OCL-TIMER_LIBRARIES})
```
The INSTALL_RPATH of `libfoo.so` would not contain the ocl library dir `<prefix>/lib/orocos/<target>/ocl`.

Both commits, b4ad7f9 and 15970d7 solve this issue in different ways. Commit c07b81b is already in master and only has been cherry-picked to this branch for testing and to avoid later merge conflicts.

It should be noted that
```cmake
orocos_use_package(ocl-logging)
orocos_use_package(ocl-timer)
orocos_component(foo foo.cpp)
```
is the recommended way to link OROCOS targets to libraries in OCL (or other OROCOS packages) which is not affected by RPATH issues.


snrkiwi:
I find it odd/alarming that this would hard code the OCL paths inside RTT. That dependency is backwards.

So what is the root difference between
```
find_package(...)
orocos_component(...)
target_link_libraries(...)
```
and
```
orocos_use_package(...)
orocos_component(...)
```
in terms of CMake code under the hood? What is the first one not doing that the second is not?

Also, how does the ORO_INSTALL_RPATH_USE_LINK_PATH option get used in the first case above? BTW where is that documented in RTT? Along with that, is there a central location that all `ORO_xxx` options are documented?

Does the same problem occur with `orocos_executable()` as well?

meyerj:
> I find it odd/alarming that this would hard code the OCL paths inside RTT. That dependency is backwards.

There are no OCL paths hardcoded inside RTT. The only effect of https://github.com/orocos-toolchain/rtt/commit/b4ad7f93d15151916718cf7d025293fb5b9c2ebe is that paths in `OROCOS-OCL_LIBRARY_DIRS` are added to the INSTALL_RPATH, but only if OCL has been found by the `find_package(OROCOS-OCL REQUIRED ...)` command, which sets this cmake variable. RTT itself and the macros shipped with RTT do not try to find OCL nor do they depend on OCL.

> So what is the root difference between [...] and [...] in terms of CMake code under the hood? What is the first one not doing that the second is not?

There is no special magic in the first cmake code snippet:

1. `find_package(OROCOS-OCL REQUIRED COMPONENTS ...)`

   - finds OCL and sets the `OROCOS-OCL_INCLUDE_DIRS`, `OROCOS-OCL_LIBRARY_DIRS`, `OROCOS-OCL_LIBRARIES` variables
   - uses the `orocos_find_package()` macro defined in rtt's [UseOROCOS-RTT-helpers.cmake](https://github.com/orocos-toolchain/rtt/blob/master/UseOROCOS-RTT-helpers.cmake#L112) under the hood
   - see https://github.com/orocos-toolchain/ocl/blob/master/orocos-ocl-config.cmake

2. `orocos_component(...)`

   - Most `orocos_*()` macros that generate a target are basically a shortcut for `add_library()`/`add_executable()` which additionally...
     - append the RTT include directories for this target only (with #85)
     - link to the RTT libraries found by `find_package(OROCOS-RTT COMPONENTS ...)` and add RTT compiler and linker flags
     - set the correct library output directory in case of catkin or rosbuild builds
     - set the INSTALL_RPATH property to the list of folders specified in the [orocos_set_install_rpath()](https://github.com/orocos-toolchain/rtt/commit/b4ad7f93d15151916718cf7d025293fb5b9c2ebe#diff-32e94bf0e0e0e2c3419f4c4e7b16ac4aL386) helper function
     - install the library/executable to the correct destination folder
     - keep a list of Orocos targets and library dirs for `orocos_generate_package()`
     - add include directories, link to libraries (unless OROCOS_NO_AUTO_LINKING is true) and add the library dirs of all packages that have been found by `orocos_use_package()` before (none in this case)

3. `target_link_libraries(...)`

   - Only required for libraries other than...
      - `OROCOS-RTT_LIBRARIES` as found by `find_package(OROCOS-RTT)`
      - libraries in packages that have been found by `orocos_use_package()` if auto-linking is enabled (none in this case)
   - The OCL libraries `OROCOS-OCL_LIBRARIES` would have to be added here explicitly, if required.
   - It is the user's responsability to place/install the linked libraries in the dynamic linker search paths or to modify the INSTALL_RPATH of the target without overriding the INSTALL_RPATH set by the `orocos_component()` macro before. My second commit, https://github.com/orocos-toolchain/rtt/commit/15970d74d3c1b8472872ed122e4fcf8ff7f9389b, relaxes this requirement a bit by restoring the pre-2.8 behavior of setting the INSTALL_RPATH_USE_LINK_PATH property for Orocos targets, but this is not a 100% solution because the library dirs at build time can be different from the final installed location in some build environments.

In the second snippet the target would be automatically linked to all targets defined by the used packages, like `ocl-logging` and `ocl-timer` in my original example, without the need for additional `find_package()`, `include_directories()` or `target_link_libraries()` commands, but also without the possibility to limit the effect to only some of the Orocos targets defined after the `orocos_use_package()` command within the same scope or a subdirectory. The `orocos_use_package()` commands have no direct effect anymore on non-Orocos targets since #85. The auto-linking feature can be disabled by setting `OROCOS_NO_AUTO_LINKING` to true.

For catkin users it should be noted that all build dependencies in the `package.xml` file will be automatically considered as Orocos package names and used if they are recognized as Orocos packages. This is usually the case if `orocos_generate_package()` has been called in the top-level `CMakeLists.txt` file.

> Also, how does the ORO_INSTALL_RPATH_USE_LINK_PATH option get used in the first case above?

All directories in the linker search path and outside the project will be added to the INSTALL_RPATH of the target (see https://cmake.org/cmake/help/v3.0/prop_tgt/INSTALL_RPATH_USE_LINK_PATH.html or https://cmake.org/Wiki/CMake_RPATH_handling). ORO_INSTALL_RPATH_USE_LINK_PATH basically replaces CMAKE_INSTALL_RPATH_USE_LINK_PATH for Orocos targets. Setting this to true resolves RPATH issues when linking to Orocos libraries found by `find_package(...)`, but only if those packages have already been installed and found at the correct location before the current build step.

It would be so much easier if Orocos would install all libraries to the default destinations without the `lib/orocos/target` hierarchy and discover packages and libraries in another way, like an XML package manifest for the installed package.

> BTW where is that documented in RTT? Along with that, is there a central location that all ORO_xxx options are documented?

There is some documentation in the headers and bodies of the [orocos-rtt-config.cmake](https://github.com/orocos-toolchain/rtt/blob/master/orocos-rtt-config.cmake), [UseOROCOS-RTT.cmake](https://github.com/orocos-toolchain/rtt/blob/master/UseOROCOS-RTT.cmake) and  [orocos-ocl-config.cmake](https://github.com/orocos-toolchain/ocl/blob/master/orocos-ocl-config.cmake) scripts, but the lack of visible and up-to-date documentation is definitely a weak point and the subject of many questions on Orocos mailing lists in the past. Take my descriptions above as a starting point. We certainly have to improve this in one way or another.

> Does the same problem occur with orocos_executable() as well?

Yes, the behavior of all macros should be more or less the same except for destination folders and some special compiler flags.

smits:
lgtm

meyerj:
Created an issue for cmake documentation here: https://github.com/orocos-toolchain/rtt/issues/165

PR #158, opened 2016-07-04
Added method setDataSource() and assignment operator to the PropertyBase interface

meyerj:
These new methods can be used to "link" a property to another without having to know the exact derived type, e.g.:

```cpp
bool linkProperty(RTT::base::PropertyBase *target,
                  const RTT::base::PropertyBase *source) {
    return target->setDataSource(source->getDataSource());
    // or
    *target = source;
    return target.ready();
}
```

```cpp
// create another property b which mirrors a
RTT::Property<double> a("a", "example property", 5.0);
RTT::PropertyBase *b = a->create();
*b = a;  // or linkProperty(b, &a)
assert(b.getDataSource() == a.getDataSource());
```

It should be noted that the assignment `*b = a` in the above example is equivalent to `b = a->copy()` because `PropertyBase::copy()` introduced in 44b9970e9685b9b812ddce1e8cc8c9b400553d54 also has linking semantics (shared data source). This is in fact inconsistent with the `PropertyBase::copy(const PropertyBase *)` or `DataSourceBase::copy()` methods which have deep-copy semantics -- so probably a candidate for removal in 2.9? The only usage of `PropertyBase::copy()` within RTT itself is [PropertyBag assignment](https://github.com/orocos-toolchain/rtt/blob/a37947a06a448dd9d8a8376e67ff1e5c8266e354/rtt/PropertyBag.cpp#L227), which is supposed to have linking semantics, too, and which has a slightly different behavior for owned vs. non-owned properties.

The `RTT::Property<T>::Property(base::PropertyBase *)` constructor and assignment operator have been updated to call the new method for consistency reasons. The patch is not ABI compatible. That's why I would like to integrate it before an official 2.9 release.

We would need this feature to build an application-wide, consolidated property tree which also allows for dynamic updates (with manual coordination).

### DIFF
--- a/rtt/Property.hpp
+++ b/rtt/Property.hpp
@@ -140,15 +140,18 @@ namespace RTT
          * @see ready() to inspect if the creation succeeded.
          */
         Property( base::PropertyBase* source)
-            : base::PropertyBase(source ? source->getName() : "", source ? source->getDescription() : ""),
-              _value( source ? internal::AssignableDataSource<DataSourceType>::narrow(source->getDataSource().get() ) : 0 )
+            : base::PropertyBase(source ? source->getName() : "", source ? source->getDescription() : "")
         {
-            if ( source && ! _value ) {
-                log(Error) <<"Can not initialize Property from "<<source->getName() <<": ";
-                if ( source->getDataSource() )
-                    log() << "incompatible type ( destination type: "<< getType() << ", source type: "<< source->getDataSource()->getTypeName() << ")."<<endlog();
-                else
-                    log() << "source Property was not ready."<<endlog();
+            if ( source ) {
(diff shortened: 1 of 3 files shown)


adolfo-rt:
LGTM modulo a small testcase that exercises the two new `PropertyBase` interface additions.

Thanks for putting this together :)

meyerj:
> LGTM modulo a small testcase that exercises the two new PropertyBase interface additions.

Done: https://github.com/orocos-toolchain/rtt/pull/158/commits/45c7a64f512eea96f44a1efaca21ec090af79f64

smits:
lgtm, I created a separate issue to track the unrelated failing test #164

PR #159, opened 2016-07-05
property: remove copy() method and allow to construct or create a property with a custom data source

meyerj:
Following my suggestion in #158, this patch removes the `RTT::base::PropertyBase::copy()` method and introduces

[PropertyBag assignment](https://github.com/orocos-toolchain/rtt/blob/953c0ec5960cf148ed78b784a5767b78c5254044/rtt/PropertyBag.cpp#L227) is supposed to have linking semantics (changing the value of a property in the copy also changes the value in the original bag), which is important for property composition and decomposition.

The `PropertyBase::copy()` method introduced in 44b9970e9685b9b812ddce1e8cc8c9b400553d54 had this linking semantics and was used for PropertyBag assignment, but the term `copy` was actually always associated with deep-copy semantics in RTT, like in PropertyBase::copy(const PropertyBase *) or in DataSourceBase::copy().

This patch introduces an alternative signature for the create() method, which accepts a custom data source instead of creating a default-initialized ValueDataSource. PropertyBag assignment now calls this `PropertyBase::create(const DataSourceBase::shared_ptr&)` method with the DataSource of the original property to be linked.

This reverts commit 44b9970e9685b9b812ddce1e8cc8c9b400553d54.

### DIFF
--- a/rtt/Property.hpp
+++ b/rtt/Property.hpp
@@ -162,7 +162,7 @@ namespace RTT
          * @post ready() will be true if datasource is a valid pointer.
          */
         Property(const std::string& name, const std::string& description,
-                 typename internal::AssignableDataSource<DataSourceType>::shared_ptr datasource )
+                 const typename internal::AssignableDataSource<DataSourceType>::shared_ptr& datasource )
             : base::PropertyBase(name, description), _value( datasource )
         {
             // need to do this on the datasource in order to have access to set()/rvalue() of the data source.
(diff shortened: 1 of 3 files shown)


smits:
We might get in trouble if someone started to use the copy method in the meantime.  But since it was broken anyway merging this is ok, they'll get an "in the face" suggestion to replace/remove that usage.  It might be wise to mention this removal in the release notes.

lgtm

meyerj:
Done: https://github.com/orocos-toolchain/orocos_toolchain/commit/196677ef8183045dcb6296c6605e873b5d1252fe

meyerj:
Good to merge?

PR #162, opened 2016-07-25
Added option, to load typekits on demand

jmachowinski:
This commit adds the option, to load typekits on demand.

This is especially useful, if one uses RTT::corba::TaskContexProxy,
as the ports are not reported, if no typekit for them is present.

### DIFF
--- a/rtt/types/TypeInfoRepository.cpp
+++ b/rtt/types/TypeInfoRepository.cpp
@@ -72,8 +72,13 @@ namespace RTT
     void TypeInfoRepository::Release() {
         typerepos.reset();
     }
-
-    TypeInfo* TypeInfoRepository::type( const std::string& name ) const
+    
+    void TypeInfoRepository::setAutoLoader(const boost::function<bool (const std::string &)> &loader)
+    {
+        loadTypeKitForName = loader;
+    }
+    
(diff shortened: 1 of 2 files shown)


jmachowinski:
Any one ? If not I assume this is ok and merge.

meyerj:
I can see how this could be useful, but the function or functor that implements the auto loading has to be written manually. Do you have an example how you would use this new interface?

jmachowinski:
A concrete usecase can be found in 
https://github.com/rock-cpp/rock-display/blob/master/src/Main.cpp
There I implemented a functor, that utilizes the fact, that 
orogen generated typekits provide a list of 'contained' types on disk.

smits:
It's definitely not related to this PR, but we should investigate the failed build:
https://travis-ci.org/orocos-toolchain/rtt/builds/147143467#L2127

lgtm

meyerj:
> It's definitely not related to this PR, but we should investigate the failed build:
https://travis-ci.org/orocos-toolchain/rtt/builds/147143467#L2127

I already pushed a patch to master: https://github.com/orocos-toolchain/rtt/commit/9140db41cb326dbf86fdd8839f9f7df8aa7047bb, but it's not yet included in this PR. I am not completely sure whether it fixes the issue on Travis.

jmachowinski:
So, merge ?

PR #163, opened 2016-08-05
Fixed decomposition of PropertyBags which contain types that implement the TemplateCompositionFactory interface

meyerj:
The method `types::decomposePropertyBag()` did not retain the type of the PropertyBag returned by `types::TypeInfo::decomposeType()` for each individual property and therefore the subsequent composition did not recognize the original type. For example, the KDL typekit implements composition and decomposition using that method (in [kdlTypekit.hpp:163](https://github.com/orocos/rtt_geometry/blob/indigo-devel/kdl_typekit/src/kdlTypekit.hpp#L163) and [motionproperties.cpp:380ff.](https://github.com/orocos/rtt_geometry/blob/indigo-devel/kdl_typekit/src/motionproperties.cpp#L380)). The fallback method of property decomposition implemented in `types::propertyDecomposition()` using `getMember()` is not affected by this issue and sets the type of the target bag correctly.

The updated test in RTT cannot easily cover this case, because no RTT core typekit implements the TemplateCompositionFactory interface.


smits:
lgtm

PR #169, opened 2016-08-12
corba: use oneway for signalling the remote on pull channels

doudou:
Signal order does not matter, so we can safely use oneway.

This "unblocks" the CORBA dispatchers on pull connections, something I always assumed had been fixed a while ago (I remember a patchset to avoid signalling pull connections that are not going to an event port). The latter would have been a better solution, but this is really a low-hanging fruit with great effect.

### DIFF
--- a/rtt/transports/corba/DataFlow.idl
+++ b/rtt/transports/corba/DataFlow.idl
@@ -79,7 +79,7 @@ module RTT
          * is available for reading.
          * @return false if this channel became invalidated.
          */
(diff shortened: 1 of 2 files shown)


meyerj:
+1, especially because it is compatible with pending PR #123.

doudou:
@meyerj: what's the merge "policy" on this kind of things ? I'm asking since you obviously reviewed the patch...

Issue #141, opened 2016-03-11
Segfault after periodic activity deletion


pdima:
I have a task with periodic activity created in updateHook():

```
bool Task::configureHook()
{
    setActivity(new RTT::Activity(ORO_SCHED_RT, m_cfg.priority, m_cfg.period));
    return true;
}

task.start();
....
task.stop(); task.configure(); task.start()
```

It always segfaults, looks like the activity loop continues to use the old activity data, deleted when the new Activity created. I tested with the current toolchain 2.9, but can reproduce the it with older versions as well.

Valgrind output:

```

==27410== Thread 7:
==27410== Invalid read of size 8
==27410==    at 0x6266C1C: RTT::os::MutexLock::MutexLock(RTT::os::MutexInterface&) (MutexLock.hpp:63)
==27410==    by 0x795865B: RTT::Activity::loop() (Activity.cpp:211)
==27410==    by 0x79681B0: RTT::os::thread_function(void*) (Thread.cpp:100)
==27410==    by 0x796CA0F: RTT::os::rtos_posix_thread_wrapper(void*) (fosi_internal.cpp:99)
==27410==    by 0x96F4181: start_thread (pthread_create.c:312)
==27410==    by 0xA22447C: clone (clone.S:111)
==27410==  Address 0xe625180 is 256 bytes inside a block of size 368 free'd
==27410==    at 0x4C2C2BC: operator delete(void*) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==27410==    by 0x7958267: RTT::Activity::~Activity() (Activity.cpp:110)
==27410==    by 0x79051DD: void boost::checked_delete<RTT::base::ActivityInterface>(RTT::base::ActivityInterface*) (checked_delete.hpp:34)
==27410==    by 0x7918547: boost::detail::sp_counted_impl_p<RTT::base::ActivityInterface>::dispose() (sp_counted_impl.hpp:78)
==27410==    by 0x5FA7301: boost::detail::sp_counted_base::release() (sp_counted_base_gcc_x86.hpp:146)
==27410==    by 0x5FA73C4: boost::detail::shared_count::~shared_count() (shared_count.hpp:371)
==27410==    by 0x78FD6E9: boost::shared_ptr<RTT::base::ActivityInterface>::~shared_ptr() (shared_ptr.hpp:328)
[...]

Issue #170, opened 2016-09-01
thread safety may be broken for OwnThread operations if sequential activity used


pdima:
I found the case when the promise is broken of OwnThread operations to be called in the thread safe way.

I have a task with sequential activity with OwnThread operations and event ports.

I found I can call OwnThread operations from task browser in the middle of update hook processing data from event port. I'd expect OwnThread operations to be serialized and called only after updateHook returned.

It's caused by SequentialActivity::thread() returning os::MainThread, so when operation is called from the main thread, OperationCallerInterface::isSend() returns false as the current thread matches activity thread even while activity is being triggered from another thread.

I don't know what would be the appropriate fix for this, OperationCallerInterface::isSend() should check the actual active sequential activity thread instead of the default MainThread, but I don't know how to do this in a thread safe way.

Another possible not very nice workarounds:
1) use a dummy thread for sequential activities so it never matches calling thread
2) OperationCallerInterface::isSend() should return true for OwnThread operations of main thread activities

meyerj:
Yes, I see the problem. This affects the [toolchain-2.9](https://github.com/orocos-toolchain/rtt/tree/toolchain-2.9) branch only due to the patch introduced in https://github.com/orocos-toolchain/rtt/commit/f5e1d78ed899b09dbdd933c2072ca25f1fd209d8.

I remember that while working on the patch we had a variant where [RTT::base::ActivityInterface::thread()](http://orocos-toolchain.github.io/devel/documentation/rtt/toolchain-2.9/api/html/classRTT_1_1base_1_1ActivityInterface.html#a190329785c1c7e4a542db833402a7811) was allowed to return a null pointer for all activities where the executing thread cannot be known by design, like for `SequentialActivity` and `SlaveActivity` without a master. `isSend()` would always return true in that case. I am not sure anymore why this approach has been disregarded at the end. Returning `os::MainThread` in these cases is just a guess, which might be valid for unit tests or if the operation is called from the TaskBrowser, but not in most other applications.

PR #171, opened 2016-09-05
Prevent changing the activity from a component's own thread

meyerj:
This patch partially solves issue #141. It is not possible to change the activity of a component from within the component's thread because it would have to destroy itself afterwards. `RTT::TaskContext::setActivity()` logs an error and returns false in this case. `RTT::TaskContext::forceActivity()` logs a fatal error.

The debug log output during thread termination has been moved from the thread destructor to the `terminate()` method to fix logging during the destruction of `RTT::Activity` (see also 58cb4bb).


meyerj:
This patch breaks some tests that call `setActivity()` on a component which returns `MainThread::Instance()` from `getActivity()->thread()`. See #170 for a related issue that is also caused by not knowing which thread will execute an activity in some cases. I am currently working on a patch based on the idea that `RTT::base::ActivityInterface::thread()` should return null in those cases.

rtt/TaskContext.cpp
```diff
@@ -344,15 +344,26 @@ namespace RTT
...
         if (this->isRunning())
             return false;
-        if ( new_act == 0) {
+
+        // refuse to setActivity from our own active thread
+        if (our_act) {
+            if (our_act->isActive() && our_act->thread() && our_act->thread()->isSelf()) {
+                log(Error) << "Cannot set the activity of TaskContext "
+                           << this->getName() << " from its own thread." << endlog();
+                delete new_act;
```
doudou: `our_act` is a shared_ptr. I'm not really sure what are supposed to be the ownership rules here.

In any case, whatever you believe the rules are, please document `setActivity` accordingly. Including and especially about this new change.

meyerj: You are right. The intended behavior is described in [TaskContext.hpp:130-143](https://github.com/orocos-toolchain/rtt/blob/a5a8fd73d5ee3522cc7a65d0167787f5218acae9/rtt/TaskContext.hpp#L130-L143):
```cpp
        /**
         * Sets the activity of this TaskContext. The
         * activity is owned by the TaskContext and you should
         * not hold a pointer to it after this call. Use
         * getActivity() lateron to retrieve a safe pointer to it.
         * @param new_act The new activity for this TaskContext,
         * which becomes owned by this TaskContext, in case this method
         * returns true.
         * @return false if this->isRunning(). You can not change the
         * activity of a TaskContext once it is running. In that case,
         * new_act is not destroyed.
         * @note This function may not be called from the current
         * ExecutionEngine thread (OwnThread), another thread (ClientThread) must call this function.
         */
```

So in case of an error (like here) the method should not destroy the activity and it is up to the caller to check the return value and eventually delete the instance.

I will remove the `delete new_act` to comply with the documentation of `setActivity()`.

meyerj: Done in https://github.com/orocos-toolchain/rtt/pull/171/commits/0e117bd6675f51776f4cc8d36bfa43e84348dbba.

PR #174, opened 2016-09-05
corba: removed special cases that depend on the data source type in CorbaTemplateProtocol<T>

meyerj:
This patch avoids some unnecessary copies during CORBA port write and read calls and during calling, sending or collecting operations. Only some of these callers of the CorbaTypeTransporter implementation actually passed instances of `LateReferenceDataSource<T>` or `LateConstReferenceDataSource<T>`.
There is no reason to differentiate between different data source types here. The generic `DataSource` and `AssignableDataSource` API should be just fine.

See also related commit 2068263a0c8a32f2e372a00dff2ad60b837f7c71.


doudou:
:+1:

PR #175, opened 2016-09-08
omit linker flag  in presence of -fsanitize

rhaschke:
In gcc environment orocos is compiled with linker option `-z defs` to ensure that all symbols in linked object files are defined. This is perfectly fine.
However, when used with [`AddressSanitizer`](https://github.com/google/sanitizers/wiki/AddressSanitizer) (option -fsanitize=address), the linker fails with lots of undefined symbols, because gcc/ld don't link shared libraries against `libasan`. This is only done for executables.

Hence, to enable linking, the linker option `-z defs` needs to be omitted when address sanitizer is used.
This patch solved the issue for me. However, checking `CMAKE_CXX_FLAGS` might not be the best option. Alternatively, I considered `CMAKE_SHARED_LINKER_FLAGS`. However, in both case there are several version of these variables depending on the choosen `CMAKE_BUILD_TYPE`.

### DIFF
--- a/UseOROCOS-RTT.cmake
+++ b/UseOROCOS-RTT.cmake
@@ -68,7 +68,9 @@ if(OROCOS-RTT_FOUND AND NOT USE_OROCOS_RTT)
 
   # This is for not allowing undefined symbols when using gcc
   if (CMAKE_COMPILER_IS_GNUCXX AND NOT APPLE)
-    SET(USE_OROCOS_LDFLAGS_OTHER "-Wl,-z,defs")
+    if (NOT CMAKE_CXX_FLAGS MATCHES "-fsanitize")
+      SET(USE_OROCOS_LDFLAGS_OTHER "-Wl,-z,defs")
+    endif()
   else (CMAKE_COMPILER_IS_GNUCXX AND NOT APPLE)
     SET(USE_OROCOS_LDFLAGS_OTHER " ")
   endif (CMAKE_COMPILER_IS_GNUCXX AND NOT APPLE)


meyerj:
Is it necessary to patch `UseOROCOS-RTT.cmake` for this? I am not familiar with AddressSanitizer, but would it be an option to simply link all libraries to `libasan` once you add `-fsanitize=address`?

Something like
```cmake
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
link_libraries(-lasan)
```

There might be some differences between GCC and Clang, but the above snippet works for me with gcc-4.8 in Ubuntu 14.04.

From https://github.com/google/sanitizers/wiki/AddressSanitizer#faq:
> Q: When I link my shared library with -fsanitize=address, it fails due to some undefined ASan symbols (e.g. asan_init_v4)?
> A: Most probably you link with -Wl,-z,defs or -Wl,--no-undefined. These flags don't work with ASan unless you also use -shared-libasan (which is the default mode for GCC, but not for Clang).

rhaschke:
I just propose to take some action in orocos's cmake file if `-fsanitize=address` is present. As the `-Wl,-z,defs` option is introduced by orocos, orocos should also deal with potential conflicts. Of course adding -lasan would be feasible as well - even better, because you don't loose the symbol-checking feature.

I'm using a global setting on the command-line / in my environment to configure a whole bunch of orocos packages. It wouldn't make sense to fix the conflict in all packages individually.

rhaschke:
Closing, because there is no more communication.
I'm using a local patch now.

PR #176, opened 2016-09-08
Do not abort if pthread_setname_np fails

meyerj:
See http://www.orocos.org/forum/orocos/orocos-users/orocosusers-getting-permission-denied-deployers-after-running-setcap-soem.

Fixed regression from bc05dceed2f123c35893bf0033d7257373f2f605.
A failure the set the name of the thread should not lead to the termination of the process.

### DIFF
--- a/rtt/os/gnulinux/fosi_internal.cpp
+++ b/rtt/os/gnulinux/fosi_internal.cpp
@@ -170,21 +170,21 @@ namespace RTT
             if (thread_name_len > MAX_THREAD_NAME_SIZE) {
                 thread_name += thread_name_len - MAX_THREAD_NAME_SIZE;
             }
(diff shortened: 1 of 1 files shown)


smits:
lgtm, thanks for fixing

Timple:
Thank you for the fix, but if this is not an error, should we not throw a warning instead of an error?

meyerj:
I assume you mean the log level of the message being printed/logged. I agree that at least for the thread names a warning is more than enough. In the second touched case, setting the CPU affinity of the created thread, an error is more appropriate and we should eventually even delete the thread, return with an error code and abort deployment. If the user assigned a non-default CPU affinity, then there was probably a reason for it.

PR #180, opened 2016-11-21
Fixed potential memory allocation problem in ConnectionManager

meyerj:
The ConnectionManager is responsible for selecting the next channel to read if an InputPort has multiple connections, which are managed in a list of `ChannelDescriptor` instances. The `ChannelDescriptor` contains the ConnPolicy of the connection, which in turn contains a `std::string` element `name_id`. Copying `ChannelDescriptor` instances is therefore not real-time safe. Luckily it is not required and we can work with a pointer (or iterator) to the current channel.

It should be noted that most implementations of `std::string` usually do not reallocate once its internal buffer is big enough to hold all potential values (all `name_id`s of connected channels). So the risk that reassigning the `std::string` element in `cur_channel` is causing heap allocations in critical paths is quite minimal. See also http://info.prelert.com/blog/cpp-stdstring-implementations.

### DIFF
--- a/rtt/internal/ConnectionManager.cpp
+++ b/rtt/internal/ConnectionManager.cpp
@@ -60,6 +60,7 @@ namespace RTT
 
         ConnectionManager::ConnectionManager(PortInterface* port)
             : mport(port)
+            , cur_channel(NULL)
         {
         }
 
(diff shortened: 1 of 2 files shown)


doudou:
Looks good to me.

Issue #164, opened 2016-08-10
Slave test failing undeterministically


smits:
I noticed in a few travis build that the slavetest sometimes failes:
 * https://travis-ci.org/orocos-toolchain/rtt/builds/147143467#L2127
 * https://travis-ci.org/orocos-toolchain/rtt/builds/142761912#L2076

meyerj:
Done (to be confirmed): https://github.com/orocos-toolchain/rtt/commit/9140db41cb326dbf86fdd8839f9f7df8aa7047bb

The above builds do not yet include this patch and I do not want to cherry-pick or merge master to not pollute the history. The slave_test failure is clearly unrelated to the respective PRs.

meyerj:
For the records: I could not reproduce the slave_test failure locally, so we can leave this issue open until there we collected some confidence that the patch fixed it on Travis.

smits:
I retriggered the jobs, since the fix is in master now I assume this resolved the build issue for the PR build. So we can close it. Or do you think this is not sufficient?

meyerj:
Travis now terminates with
```
25: Test command: /home/travis/build/orocos-toolchain/rtt/build/tests/slave_test
25: Test timeout computed to be: 1500
25: Running 2 test cases...
25: /home/travis/build/orocos-toolchain/rtt/tests/slave_test.cpp(154): error in "testSlaveOperationCall": check handle.collectIfDone() failed

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

So apparently the patch turned the failure into a dead-lock, while for master itself the tests exit successfully, e.g. https://travis-ci.org/orocos-toolchain/rtt/builds/151199255.

Not sure if Travis really tests the merged commit if you retrigger it.

meyerj:
Unfortunately this still seems to be an issue. I just observed this error and dead-lock locally when running unit tests for toolchain-2.9, which includes the proposed patch. For a second invocation the slave_test passed without issues.

PR #181, opened 2016-11-22
mqueue: enforce MQ name convention

doudou:


rtt/transports/mqueue/MQSendRecv.cpp
```diff
@@ -75,11 +76,14 @@ void MQSendRecv::setupStream(base::DataSourceBase::shared_ptr ds, base::PortInte
...
     mis_sender = is_sender;
 
-    std::stringstream namestr;
-    namestr << '/' << port->getInterface()->getOwner()->getName() << '.' << port->getName() << '.' << this << '@' << getpid();
-
     if (policy.name_id.empty())
-        policy.name_id = namestr.str();
+    {
+        std::stringstream name_stream;
+        name_stream << port->getInterface()->getOwner()->getName() << '.' << port->getName() << '.' << this << '@' << getpid();
```
meyerj: Not really related to this PR, but in general a port does not need to have a `DataFlowInterface` and `getInterface()` can return a null pointer. Additionally, but more unlikely, a `DataFlowInterface` does not need to be owned by a `TaskContext` and `getOwner()` can also return null. We already observed segfaults because of this at other places. These are all public and documented RTT classes and we should not assume that ports are only used inside components.

Not sure if for this special case it is guaranteed that none of the pointers is invalid, but maybe it is a good idea to also add a check together and eventually omit the component part in the MQ name?

meyerj:
One comment about missing null pointer checks, otherwise lgtm.
Build failures in Travis are unrelated - see https://github.com/orocos-toolchain/rtt/issues/164.

rtt/transports/mqueue/MQSendRecv.cpp
doudou: Good to know ... I for one always assumes that this is the case.

As you say not really related to this PR (I just copied the existing code). However, as the goal here is to create a globally unique name, maybe it would be best to throw if there is no interface ? (Forcing the user to provide a name_id in this case)

doudou: Ping @meyerj

meyerj: I am not familiar with the naming requirements of message queues as I never used that transport myself. If you say it has to be globally unique, whatever global means in this context, it might be the best option to log an error that explains the issue and throw. I leave that decision to you. But we should definitely avoid segmentation faults and I would ask you to implement whatever exception behavior you find most reasonable. Or add a new issue with a reference to this discussion so that we at least not forget about it later.

doudou: OK ... Since you leave it to me, I'll throw.

For your information "global" means "system wide" in this context.

doudou:
Updated

PR #183, opened 2016-12-16
corba: expose the CircularBuffer connection type on CORBA

doudou:


meyerj:
Note: This patch is also part of #114, but as the `CConnectionModel` enum is the same in both patches there is no conflict.

PR #184, opened 2016-12-16
mqueue: ignore EINTR in select()

doudou:


meyerj:
Looks good to me. Does the same issue also affect the select call in the [FileDescriptorActivity](https://github.com/orocos-toolchain/rtt/blob/master/rtt/extras/FileDescriptorActivity.cpp)?

doudou:
>  Does the same issue also affect the select call in the FileDescriptorActivity?

I would say so.

PR #187, opened 2016-12-23
add tracing support using lttng-ust

doudou:
This adds basic infrastructure, and the most useful trace points (in/out of hooks, read/write ports). It's actually a very old branch, initiated by Paul Chavent, that I revived two or three times over the years. I've used it extensively in the last weeks, and I'd like to finally have it in mainline.

Even if the support is compiled-in, it is not enabled unless (1) the tracing library is LD_PRELOAD'ed into the RTT binary and (2) lttng itself is enabled. (1) is critical to make the support no-cost if one has it built, but does not use it.

In order to avoid memory allocations in RTT itself on the tracing path (which is a fairly hot path), I added C-string versions of some objects (ports, tasks) that are passed to lttng on tracing. I have to admit that I don't know what are lttng's own memory allocation profile though.


snrkiwi:
Interesting work Sylvain.

1) Suggest making this a feature that you ENABLE, rather than one you do NOT DISABLE.

2) Out of curiosity, why are some definitions outside the multi-include guard in rtt/os/gnulinux/traces/lttng_ust.h?

3) How does this differ from existing facilities like kernelshark and ftrace? We've used these extensively to see what is going on within a deployment.

4) The need to have cName is a bit of a PITA, but understandable I think.

Whatever you're tracing with this it doesn't look like a lot of fun ......

doudou:
> Suggest making this a feature that you ENABLE, rather than one you do NOT DISABLE.

I was expecting one like that. I'll do it !

> Out of curiosity, why are some definitions outside the multi-include guard in rtt/os/gnulinux/traces/lttng_ust.h?

It's lttng-mumbo-jumbo. It basically ensures that the probes are generated in the tracing library, but that the rest of RTT only sees "empty hooks" that will be then replaced through LD_PRELOAD at runtime.

> How does this differ from existing facilities like kernelshark and ftrace? We've used these extensively to see what is going on within a deployment.

I honestly don't know. The RTT-lttng work is already a few year old, so it was very low-cost for me to refresh it. With the very limited and theoretical knowledge I have of kernelshark/ftrace, I would guess that the biggest difference is that lttng-ust is really geared for userspace, and allows to define custom events - so that one does not need to know about RTT's internals (resolving function names, ...) to use the tracer. It also has ready-to-use generic userspace and kernel tracing (scheduling, memory allocation, ...) that you can mix with your custom events. After quick googling, it seems that the consensus is that lttng is great for userspace, and ftrace is more designed for kernel tracing. Now, I take it with a grain of salt, it's really a quick glance at the blogosphere's opinion on the subject ;-)

Does ftrace generate ctf traces ?

> Whatever you're tracing with this it doesn't look like a lot of fun ......

It was https://github.com/orocos-toolchain/rtt/pull/182 ... So not really, no ...

Now, it did force me to add relevant support to our system manager (Syskit), which now also generates tracecompass-compatible traces (ctf). This is now pretty awesome.

rtt/base/TaskCore.cpp
```diff
@@ -337,5 +352,12 @@ namespace RTT {
         }
     }
 
+    std::string TaskCore::getName() const {
+        return name;
+    }
+    void TaskCore::setName(const std::string& name) {
+        this->name = name;
+        this->cName = strdup(name.c_str());
```
meyerj: Why does lltng-ust require this copy and we cannot simply replace all occurences of `cName` with `name.c_str()`?

rtt/base/TaskCore.cpp
```diff
@@ -53,6 +55,7 @@ namespace RTT {
            ,mTaskState(initial_state)
            ,mInitialState(initial_state)
            ,mTargetState(initial_state)
+           ,cName(strdup(""))
```
meyerj: Add a `TaskCore` constructor which directly takes a `const string &` argument, analog to [TaskContext](https://github.com/orocos-toolchain/rtt/pull/187/files#diff-a739de4d537602c73893d46c20da0a31R70)?

rtt/base/TaskCore.hpp
```diff
@@ -337,6 +337,17 @@ namespace RTT
             return ee;
         }
 
+        /**
+         * The name of this TaskCore
+         *
+         * Used for debugging and/or tracing purposes
+         */
+        std::string getName() const;
```
meyerj: Return type `const std::string &` like in [const std::string& TaskContext::getName() const](https://github.com/orocos-toolchain/rtt/blob/8e995a22ab8a545748fae84dc7095a3007edfee0/rtt/TaskContext.hpp#L128)?

rtt/base/TaskCore.hpp
```diff
@@ -337,6 +337,17 @@ namespace RTT
...
+
+        /** Sets the name of this TaskCore
+         */
+        void setName(const std::string& name);
```
meyerj: Theoretically names can become inconsistent as a TaskContext's name can be changed via `tc->setName(...)` inherited from TaskCore and also via `tc->provides()->setName(...)`, which is the return value of `tc->getName()`. Please add an override of `setName()` in class TaskContext that sets both.

Or remove the `setName()` method completely in favor of the constructor overload. Changing the name of a component after its construction does not play nicely together with full port names anyway.

rtt/base/TaskCore.cpp
doudou: Because c_str() is lifetime-sensitive, which does not play nice with the tracing. Since I needed a longer lifetime, I also preferred removing all issues related to memory allocation (since c_str() has no guarantee there either).

rtt/base/TaskCore.hpp
doudou: :+1:

rtt/base/TaskCore.hpp
doudou: :+1:

rtt/base/TaskCore.cpp
doudou: :+1:

rtt/base/TaskCore.cpp
doudou: I have to admit that I jumped a bit to conclusion here ... I did see issues with the name being garbage in the trace and attributed it to c_str ... But I'm not so sure anymore.

doudou:
@meyerj I've implemented your suggestions regarding the naming on TaskCore. If you prefer, I can hold on to this PR until I have a better idea about the C-string thing (but that's going to take some time ...)

doudou:
Ping

doudou:
Thanks @meyerj. Your suggestions sound good. I'll have a go at them when I have the time.

The more tracing points, the merrier. So, :+1: to add more, but I would ask for your help there, you know this part of the codebase much better than I do. I would prefer keeping the existing hooks configure/start/stop/cleanup, because even if they're not critical for realtime analysis, they are still useful for multiprocess race analysis.

rtt/os/gnulinux/traces/lttng_ust.h
```diff
@@ -0,0 +1,98 @@
...
+    (
+        ctf_integer(unsigned, value, v)
+    )
+)
+
+#ifndef __cplusplus
+#define TRACEPOINT_CONTEXT(provider, event, ...) \
+    TRACEPOINT_EVENT(provider, event, TP_ARGS(unsigned, v, const char *, s), TP_FIELDS(ctf_string(name, s) ctf_integer(unsigned, on, v)))
+#else
+#define TRACEPOINT_CONTEXT(provider, event, ...) \
```
meyerj:
```cpp
In file included from /opt/orocos/kinetic/src/orocos_toolchain/rtt/rtt/os/traces.h:7:0,                                                                                                                                                
                 from /opt/orocos/kinetic/src/orocos_toolchain/rtt/rtt/Activity.cpp:46:                                                                                                                                                
/opt/orocos/kinetic/src/orocos_toolchain/rtt/rtt/os/gnulinux/traces/lttng_ust.h:82:57: warning: ISO C++11 requires at least one argument for the "..." in a variadic macro                                                             
 TRACEPOINT_CONTEXT(orocos_rtt, TaskContext_configureHook)                                                                                                                                                                             
                                                         ^
```
and many more.

Simply remove the `...`? I cannot find any usage of additional arguments.

rtt/base/TaskCore.hpp
meyerj: I found that the CORBA `TaskContextProxy` changes the name of the underlying `TaskContext` after initial construction in [TaskContextProxy.cpp:151](https://github.com/orocos-toolchain/rtt/blob/48279d7539f26bf344f905ea411e6e6067db1a70/rtt/transports/corba/TaskContextProxy.cpp#L151) or [TaskContextProxy.cpp:182](https://github.com/orocos-toolchain/rtt/blob/48279d7539f26bf344f905ea411e6e6067db1a70/rtt/transports/corba/TaskContextProxy.cpp#L182).

Not a big issue because proxies are not supposed to be traced, but they do create (and immediately destroy) a thread and trigger the `thread_init` tracepoint.

rtt/os/gnulinux/traces/lttng_ust.h
meyerj: Confirmed that removing the unused full stops would work.

francisco-miguel-almeida:
> I could successfully trace the execution of a unit test using this patch. It might require some documentation on how to use it (with `LD_PRELOAD`).

The feature indeed needs to be documented, some explanation definitely will need to be added in the changelog for the release where LTTng support is introduced.
 
> LGTM after the following minor comments have been addressed.

Removed the variadics from the macros, confirmed on my end as well. This being the only comment that needed to be addressed, I can now merge to `master`.

PR #190, opened 2017-01-04
logger: Ensure that the notification of real-time logging being enabl…

snrkiwi:
The notification that real-time logging was enabled can be lost due to the sequence of "log then enable". The fix is to "enable then log", which ensures that the notification comes out if the log level is currently RealTime.


meyerj:
lgtm

Issue #166, opened 2016-08-10
Remove support for gcc versions without visibility support


meyerj:
From https://github.com/orocos-toolchain/rtt/pull/154#issuecomment-223905893:

> Could we remove the gcc version check completely? Does anyone still uses a gcc version older than 3.3, where the visibility support has been added?
> 
> If the compiler check in CMakeLists.txt would be removed, the value of the RTT_GCC_HASVISIBILITY cmake variable and preprocessor macro must be considered as true and removed from the generated header files.

meyerj:
The visibility attributes are completely unused at the moment anyway because RTT is not built with `-fvisibility=hidden` by default. The corresponding lines in cmake config have been disabled in https://github.com/orocos-toolchain/rtt/commit/72bc0bd1b32e4723c6000512e43de218bfbfa9f2 and mostly removed since then. The only commented leftover is in [rtt/typekit/CMakeLists.txt:17](https://github.com/orocos-toolchain/rtt/blob/9c46c957bcd6078995f4d36ebeec4da385ed8218/rtt/typekit/CMakeLists.txt#L17). Might have been commented and committed accidentally back then?

PR #191, opened 2017-01-05
cmake gcc6 support

planthaber:


meyerj:
That reminds me at #166: Should we remove the version check completely? Visibility support can be considered as standard nowadays. It was added in GCC 3.3.

planthaber:
But https://github.com/orocos-toolchain/rtt/issues/166  is independent of this PR, is it?
This fixes compilation with gcc6 in the current state.
The check can be removed later if that is the decision of https://github.com/orocos-toolchain/rtt/issues/166

meyerj:
Agreed. But a simple removal of the whole version check would IMHO also solve the problem and saves us from updating the check with every new gcc major version. Even if we keep the check, it would be possible to rewrite it to accept any newer version.

PR #192, opened 2017-01-05
boost 1.61 support

planthaber:
Partial Function Template Ordering was removed from boost in 1.59, the according define BOOST_PFTO also.

I am not really sure if the define is the correct way to solve it, bur at lease the functionality doesn't change for users with older boost versions

Also changed some ifdefs, from checking against 1.62 to 1.61 (current default in ubuntu 16.10)

I wonder how it compiled against 1.62 before, somebody might have done it as the boost::serialization::array_wrapper defines were already here

### DIFF
--- a/rtt/transports/mqueue/binary_data_archive.hpp
+++ b/rtt/transports/mqueue/binary_data_archive.hpp
@@ -85,9 +85,9 @@
 #include <boost/serialization/item_version_type.hpp>
 #endif
 
-#if BOOST_VERSION >= 105900
+#ifndef BOOST_PFTO
 //Partial Function Template Ordering removed from boost in 1.59, setting it to nothing seems to work
-//(was also done in <boost/ptfo.hpp> itself)
+//(was also done in <boost/ptfo.hpp> itself, header removed from 1.59)
 //and keeps its functionality for older boost versions
 #define BOOST_PFTO
 #endif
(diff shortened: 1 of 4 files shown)


meyerj:
The guard for the renamed `boost::serialization::array_wrapper<T>` template class has been introduces recently in https://github.com/orocos-toolchain/rtt/commit/87659188c71bf3013247590a386321648dd8fb67. You are right that the check should have been against Boost 1.61 instead. Could you please also update the remaining check in [mqueue_archive_test.cpp:111](https://github.com/planthaber/rtt/blob/0d83c908a6f06934cbf6b872537a49edcaf730fd/tests/mqueue_archive_test.cpp#L111)?

We tested the patch for the Max OS X target with Boost 1.63 and Xcode 7.3 and 8.2 in Travis: https://travis-ci.org/orocos-toolchain/orocos_toolchain/builds/187551156. There was no error due to the undefined `BOOST_PFTO` macro for some reason.

rtt/transports/mqueue/binary_data_archive.hpp
```diff
@@ -85,6 +85,13 @@
 #include <boost/serialization/item_version_type.hpp>
 #endif
 
+#if BOOST_VERSION >= 105900
```
meyerj: Maybe we can replace this check by
```cpp
#ifndef BOOST_PFTO
```
? That way there is no error if the macro would already be defined elsewhere, e.g. if the `boost/ptfo.hpp` is included elsewhere.

planthaber:
Updated the PR

The boost/ptfo.hpp header is completely removed from boost.
Anyways checking #ifndef BOOST_PFTO is still the better solution.

Also fixed the test case guard for mqueue_archive_test.cpp:111

meyerj:
Thanks!

Issue #179, opened 2016-11-19
Service destructor is not called when component is unloaded


disRecord:
This behavior can be reproduced by simple test

```
    class MyService : public RTT::Service { 
    public:
        MyService(TaskContext* owner) : Service("myservice", owner)  {
             cout << "MyService constructor." << endl;    
        }

       ~MyService() {
             cout << "MyService destructor." << endl;
        }
    };
```

and then load `myservice` in any component and shutdown deployer.

It seems to be caused by boost::shared_ptr loop in RTT::Service definition (http://github.com/orocos-toolchain/rtt/blob/master/rtt/Service.hpp#L576):

```
     typedef std::map< std::string, shared_ptr > Services;
     Services services;
     ...
     shared_ptr parent;
```

meyerj:
We already had similar problems with other services, e.g. for the `rosservice` service (see https://github.com/orocos/rtt_ros_integration/pull/60). It would be certainly better to use weak pointers for the parent to break the loop, but normally destruction works fine because the TaskContext destructor actively clears the tasks services in [TaskContext.cpp:135](https://github.com/orocos-toolchain/rtt/blob/master/rtt/TaskContext.cpp#L135).

However, there are many reasons why other shared pointers to a service might still exist and the service is not destroyed, e.g. when there is a CORBA TaskContextProxy.

disRecord:
Excuse me, It seems I missed important information in original post: this issue is related with `toolchain-2.9` version of OROCOS. Code in [TaskContext.cpp:~TaskContext.cpp]](../blob/toolchain-2.9/rtt/TaskContext.cpp#L118) is entirely different here.

meyerj:
Thanks for the detailed analysis!

> 1. Method `clear()` in [Service.cpp:219](../blob/master/rtt/Service.cpp#L219) is not recursive. `removeService()` is called for each subservice, but it simply reset shared pointers and does not call `clear() on corresponding Services. So if any of subservices has its own subservice we have `shared_ptr` loop.

You are right that this might be an issue in some scenarios, but making `clear()` recursive might break existing use cases and is not a good option.

> 2. Another possible cause is  `shared_from_raw()`: [Service.hpp:109](../blob/master/rtt/Service.hpp#L109). It seems there are many pitfalls related with this method: http://stackoverflow.com/questions/22184772/why-is-boostenable-shared-from-raw-so-undocumented

[...]

PR #193, opened 2017-01-05
corba: manage links to Service instances in weak pointers

meyerj:
The CORBA layer should not delay the destruction of Service instances once the owning TaskContext gets destroyed. Accessing a destroyed service instance through CORBA would result in a bad_weak_ptr exception.

Might be related to https://github.com/orocos-toolchain/rtt/issues/179.


meyerj:
Duplicate of #238, which already has been merged on Nov 13, 2017.

PR #196, opened 2017-01-06
Fixed OperationInterfacePartFused for -std=c++11 builds with Boost 1.58 (master)

meyerj:
The [Boost unfused adapter](http://www.boost.org/doc/libs/1_58_0/libs/fusion/doc/html/fusion/functional/adapters/unfused.html) accepts only LValue arguments, while in general operations can also have RValues as arguments (const references or temporaries). Without C++11 or for older Boost versions this was apparently not an issue and the necessary conversion have been done implicitly (?).

The solution applies the [forward_adapter template](http://www.boost.org/doc/libs/1_63_0/libs/functional/forward/doc/html/index.html) to overcome this limitation as suggested in the Boost Fusion documentation.

This patch also reverts the `USE_CPP11` macro usage for this file, as it did not use `std::bind` or any other C++11 feature anyway and the only effect was the namespace of placeholders (see https://github.com/orocos-toolchain/rtt/commit/35cd426b4fd93c14a597b6ae590cc45c2dd2d105). Maybe it was unintentionally that `boost::bind` was not replaced by `std::bind` in the `USE_CPP11` section of OperationInterfacePartFused.hpp like in other files of that patch (#103).
`boost::arg<i>` is a simpler solution than introducing the macro and use C++11 that also prevent namespace clashes without having to use boost::lambda.

Tested on Ubuntu Xenial with gcc 5.4, Boost 1.58 and `-std=c++11` compile flag.
[This](https://gist.github.com/meyerj/ff68336793243d58eb821696af60342c) is the build error I got without this patch.

### DIFF
--- a/rtt/internal/OperationInterfacePartFused.hpp
+++ b/rtt/internal/OperationInterfacePartFused.hpp
@@ -40,7 +40,6 @@
 #define ORO_OPERATION_INTERFACE_PART_FUSED_HPP
 
 
-#include <boost/config.hpp>
 #include <boost/shared_ptr.hpp>
 #include <boost/function_types/result_type.hpp>
 #include <boost/function_types/parameter_types.hpp>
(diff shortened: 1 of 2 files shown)


rtt/internal/OperationInterfacePartFused.hpp
```diff
@@ -171,7 +169,7 @@ namespace RTT
                 assert( carity == collectArity() + 1 ); // check for arity functions. (this is actually a compile time assert).
                 if ( args.size() != carity ) throw wrong_number_of_args_exception(carity, args.size() );
                 // we need to ask FusedMCollectDataSource what the arg types are, based on the collect signature.
-                return new FusedMCollectDataSource<Signature>( create_sequence<typename FusedMCollectDataSource<Signature>::handle_and_arg_types >::sources(args.begin()), blocking );
+                return new FusedMCollectDataSource<Signature>( create_sequence<typename FusedMCollectDataSource<Signature>::handle_and_arg_types >::assignable(args.begin()), blocking );
```
meyerj: Not really related to this PR, but shouldn't `produceCollect()` only accept a sequence of assignable data sources? Same in [line 424](https://github.com/orocos-toolchain/rtt/pull/196/files#diff-7ea54099da3c95b19cb4c6cb852b43c1R424) for class `OperationInterfacePartFusedDS`.

meyerj:
I discussed this patch with @psoetens last week and he agreed with the solution.

PR #195, opened 2017-01-05
Travis: switch to containerized build environment and add clang and C++11 jobs

meyerj:
Example build: https://travis-ci.org/orocos-toolchain/rtt/builds/189295270

Travis supports [containerized builds](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/) since 2 years, which improve build times. With the apt add-on it is also possible to install system dependencies without the need for sudo.

I also added a build matrix to test clang and gcc builds with and without c++11 enabled. We might also include Mac OS X here.

We are already running integration tests for the toolchain for the orocos_toolchain repository, but without unit tests: https://github.com/orocos-toolchain/orocos_toolchain/pull/15 (only log4cpp, rtt and ocl)

PR #198, opened 2017-01-10
Replaced all occurences of deprecated auto_ptr by unique_ptr

meyerj:
auto_ptr was deprecated in C++11. Removes compiler warnings when compiling with `-std=c++11`.
See http://en.cppreference.com/w/cpp/memory/auto_ptr.

### DIFF
--- a/rtt/types/PropertyDecomposition.cpp
+++ b/rtt/types/PropertyDecomposition.cpp
@@ -48,6 +48,7 @@
 #include <vector>
 #include <string>
 #include <memory>
+#include <boost/config.hpp>
 #include <boost/lexical_cast.hpp>
 #include "../Logger.hpp"
 #include "TypeInfo.hpp"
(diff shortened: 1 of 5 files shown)


snrkiwi:
So is C++11 a *requirement* now for RTT? If so, where is that documented?

meyerj:
No, it's not a requirement. With the patch RTT only uses `unique_ptr` if available (using [Boost Config macros](http://www.boost.org/doc/libs/1_63_0/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.macros_that_allow_use_of_c__11_features_with_c__03_compilers)) and falls back to `auto_ptr` otherwise. The main reason is to avoid deprecated warnings in some compilers, especially GCC 5 or Clang with C++11 enabled. See https://travis-ci.org/orocos-toolchain/rtt/jobs/190786861 for an example.

Like some of the other PRs I submitted last week this patch is a consequence of the build errors and warnings revealed by the new Travis build matrix introduced in https://github.com/orocos-toolchain/rtt/pull/195.

To our best knowledge the only PR that would raise the compiler version requirements to at least GCC 4.2 is  https://github.com/orocos-toolchain/rtt/pull/200, which removes the version check that enables or disables visibility support conditionally. But I don't think that anyone still uses older compilers or Boost versions. We recently verified successful compilation and that unit tests still pass for the current master branch in Ubuntu Lucid for https://github.com/orocos-toolchain/rtt/pull/201 (gcc 4.4, Boost 1.34).

But you are right that using Boost Config for the check raises the minimum Boost version requirement to 1.50, which introduced the `BOOST_NO_CXX11_SMART_PTR` macro. Probably checking `__cplusplus` instead is the more backwards compatible solution. This would also affect #196 and #197.

meyerj:
> But you are right that using Boost Config for the check raises the minimum Boost version requirement to 1.50, which introduced the BOOST_NO_CXX11_SMART_PTR macro. Probably checking __cplusplus instead is the more backwards compatible solution. This would also affect #196 and #197.

I replaced Boost Config macros by `__cplusplus > 199711L` checks.

PR #199, opened 2017-01-10
toolchain-2.9 backwards compatibility

meyerj:
Some minor patches to fix backwards compatibility with older CMake and Boost versions. According to the [installation manual](https://orocos-toolchain.github.io/devel/documentation/rtt/toolchain-2.9/xml/orocos-installation.html) Orocos RTT still supports CMake 2.6.3 and Boost 1.33 (1.40 recommended).

I propose to raise the minimum versions to Boost 1.40 and CMake 2.8.0. These are the versions provided in Ubuntu Lucid and RTT and OCL still compile and pass all unit tests in Lucid with these few patches.

See commit messages for some more details.

TODO:
- [ ] update installation manual


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 #
 # Test CMake version
 #
-CMAKE_MINIMUM_REQUIRED(VERSION 2.8.3)
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
```
meyerj: I was not aware that [cmake_minimum_required()](https://cmake.org/cmake/help/v3.0/command/cmake_minimum_required.html) also implicitly sets the cmake policies to the default for the specific version. I have not verified yet that this change does not have any side effects.

The affected policies are [CMP0015](https://cmake.org/cmake/help/v3.0/policy/CMP0015.html) and [CMP0016](https://cmake.org/cmake/help/v3.0/policy/CMP0016.html). I think both should have no effect on RTT itself, but we should consider to remove the line from the `UseOROCOS-RTT*.cmake` scripts or use the policy stack inside macros because users include them in their own projects and the `cmake_minimum_required()` line can have side effects.

snrkiwi:
LGTM

PR #200, opened 2017-01-10
cmake: remove gcc version check for visibility support

meyerj:
The gcc version check has been updated several times (#191, #154) in the past to support newer versions and involves a fragile compiler execution using `execute_process()`. I think it's time to simply drop support for older versions that do not yet support visibility attributes, older than gcc 4.2 released in 2007, which introduced visibility support for types.

Note that RTT does not set the default visibility to hidden on its own (by adding the `-fvisibility=hidden` option) and all symbols are visible by default. Perhaps this option was even removed/commented accidentally during the refactoring in 72bc0bd1b32e4723c6000512e43de218bfbfa9f2.

See also https://gcc.gnu.org/wiki/Visibility.

TODO:
- [ ] Update minimum gcc version in the [installation manual](https://orocos-toolchain.github.io/rtt/master/xml/orocos-installation.html) to 4.2
- [ ] Test the toolchain with the `-fvisibility=hidden` compile option
- [ ] Set default visibility to hidden for RTT?


meyerj:
#235 follows up on the visibility issue.

PR #203, opened 2017-01-17
Fix/unneeded locking in logger

smits:
ptal @meyerj, this should get rid of any undesired locking within the logger if logging is not allowed.

### DIFF
--- a/rtt/Logger.cpp
+++ b/rtt/Logger.cpp
@@ -421,6 +421,8 @@ namespace RTT
 
     Logger& Logger::in(const std::string& modname)
     {
+        if ( !d->maylog() )
+            return *this;
         os::MutexLock lock( d->inpguard );
         d->moduleptr = modname.c_str();
         return *this;
@@ -428,12 +430,16 @@ namespace RTT
 
     Logger& Logger::out(const std::string& oldmod)
     {
+        if ( !d->maylog() )
+            return *this;
         os::MutexLock lock( d->inpguard );
         d->moduleptr = oldmod.c_str();
         return *this;
     }
 
     std::string Logger::getLogModule() const {
+        if ( !d->maylog() )
+            return "";
         os::MutexLock lock( d->inpguard );
         std::string ret = d->moduleptr.c_str();
         return ret;


meyerj:
I think this patch would fix locking issues if the RTT logger is not enabled at all, so good to merge as a workaround if the Logger is completely disabled.

`maylog()` always returns true, independent of the log level, if the Logger is running, unless the log level is set to `RealTime` and real-time logging is enabled (see [Logger.cpp:157](https://github.com/orocos-toolchain/rtt/blob/a57c54bd11ae99473df26537feb3302c3e18dd30/rtt/Logger.cpp#L157)). To disable logging completely, the `ORO_LOGLEVEL` enviroment variable has to be set to a negative value. Even then then RTT Logger creates an empty `orocos.log` file, which could be considered a bug.

There might be race conditions if the log level would change between `RealTime` and other values while a thread still has an `In` instance, like if it bypassed `Logger::in(...)` but calls `Logger::out(...)` or vice-versa. But it would only have an effect on the module name output in logs, so this might be a corner case.

Peter once worked on an alternative solution where the module information is stored on the stack only:
https://github.com/orocos-toolchain/rtt/compare/master...psoetens:master-loggerupdate
This patch also fixed the broken module name printing if `Logger::In` sections from different threads overlap, an issue which actually renders this feature completely useless for a framework like RTT. Unfortunately the required syntax it not backwards-compatible. In order to achieve this, we would probably need target-independent, thread-local storage support (something like (Boost Thread Local Storage](http://www.boost.org/doc/libs/1_63_0/doc/html/thread/thread_local_storage.html)).

I think we should dig a bit deeper here and try to find a solution that
- is fully API compatible
- does not involve locks and minimizes run-time overhead if the log line is discarded
- ideally provides its own thread-local, real-time or statically allocated buffer to assemble the log lines
- fixes the `Logger::In` multi-threading issue
- does not create an empty `orocos.log` file if logging is disabled

meyerj:
We should use `os::AtomicInt` for the `started`, `allowRT` and `outloglevel` variables if they are accessed without locking in `maylog()` and other methods. However, for most platforms integer and boolean operations might be atomic, but this argument would invalidate the purpose of the `os::AtomicInt` type completely.

snrkiwi:
LGTM. I think that this fixes the fundamental issue that this pull request is about.

@meyerj if we want to throw darts at the logger system as a whole, then there's plenty of material there. But I think we should push those to other, future, pull requests.

PR #204, opened 2017-02-02
Fix service destruction

meyerj:
Fixes the service destruction bug as reported in https://github.com/orocos-toolchain/rtt/issues/179. We should clear() the tasks service and requester interface on TaskContext destruction to not prevent destruction of children unless they are still referenced from somewhere else.

Includes https://github.com/psoetens/rtt/pull/8, but adds a commit which recursively resets the owner pointer of all children in case there is still another reference. In this case they should at least know that their owning TaskContext went away.

### DIFF
--- a/rtt/Service.cpp
+++ b/rtt/Service.cpp
@@ -123,26 +123,18 @@ namespace RTT {
     }
 
     Service::shared_ptr Service::provides(const std::string& service_name) {
-        if (service_name == "this")
-            return provides();
-        shared_ptr sp = services[service_name];
+        shared_ptr sp = getService(service_name);
         if (sp)
             return sp;
         sp = boost::make_shared<Service>(service_name, mowner);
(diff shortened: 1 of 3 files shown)


rtt/Service.cpp
```diff
@@ -263,19 +264,8 @@ namespace RTT {
...
-                // we pass and store a shared ptr in setParent, so we hack it like this:
-                shared_ptr me;
-                try {
-                    me = shared_from_this();
-                } catch ( boost::bad_weak_ptr& bw ) {
-                    me.reset(this); // take ownership
-                }
-                it->second->setParent( me );
-            }
-        }
```
meyerj: Not sure why this section was ever required. When iterating over child services in `Service::setOwner()` their parent pointer should already be set to ourselves (called from `Service::addService()`), so no need to reset it here.

PR #206, opened 2017-02-16
Another fix for blocking function calls in non-periodic components

meyerj:
This would be another, much simpler solution for the blocking function call issue reported in https://github.com/orocos-toolchain/rtt/issues/150.

Previous proposals have been:
1. ~~https://github.com/psoetens/rtt/pull/7~~ (was invalid)
2. a. https://github.com/orocos-toolchain/rtt/pull/156: run called functions asynchronously, like an operation, but breaks yielding in functions being called
   b. https://github.com/meyerj/rtt/pull/6: extension of the previous, which additionally supports yields by falling back to function semantics in that case
3. The new proposal here is as follows:

Whenever a scripting function is called, we also trigger a full update cycle, as if `TaskContext::trigger()` had been called. This is a no-op for periodic activities, but otherwise makes sure that the function gets executed soon without an explicit trigger. The caveat is that it also triggers an update hook, which I think is acceptable for non-periodic components. The "best" option would be to only enqueue the function being called, but that's already close to the solution in https://github.com/orocos-toolchain/rtt/pull/156 and https://github.com/meyerj/rtt/pull/6.

Additionally, if the caller and the runner's thread are the same, functions will be processed in a loop now in `waitAndProcessFunctions()` because there is nothing else we can do anyway to resolve the situation and otherwise the engine would have to rely on another external trigger to continue and finish the current cycle. This change also effectively disables yields, but only if one function calls another within the same component. The current implementation, which blocks on the `msg_cond` condition variable, does not guarantee that functions execute at most once in an update cycle either, because a callback type trigger was sufficient to initiate another `processFunctions()` cycle.

This patch also contains a small optimization for `waitAndProcessMessages()`, which checked the predicate twice within each cycle.

The updated `scripting_test` has been copied from https://github.com/meyerj/rtt/pull/6 and even checks some more complex cases of nested function calls which yield.

### DIFF
--- a/rtt/ExecutionEngine.cpp
+++ b/rtt/ExecutionEngine.cpp
@@ -349,7 +349,10 @@ namespace RTT
     void ExecutionEngine::waitAndProcessMessages(boost::function<bool(void)> const& pred)
     {
         assert( mmaster == 0 );
-        while ( !pred() ){
+        if ( pred() )
+            return;
+
+        while ( true ){
             // may not be called while holding the msg_lock !!!
             this->processMessages();
             {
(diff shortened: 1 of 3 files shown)


meyerj:
After a discussion with @psoetens last week we will probably go for proposal 2b mixed with the `ExecutionEngine` patches in this PR. `ExecutionEngine::waitAndProcessFunctions()` will be removed completely as its semantics are broken and it has unwanted side effects. Updated patch pending...

meyerj:
https://github.com/meyerj/rtt/pull/6 and the ExecutionEngine patches of this PR have been merged into #156.

PR #208, opened 2017-02-23
Fixed memory leak in RTT::updateProperties() and RTT::marsh::CPFMarshaller::deserialize()

meyerj:
Discovered while testing some things with the [valgrind memory error checker](http://valgrind.org/docs/manual/mc-manual.html)...

### DIFF
--- a/rtt/PropertyBag.cpp
+++ b/rtt/PropertyBag.cpp
@@ -615,7 +615,7 @@ namespace RTT
                             temp->update( (*sit) );
                         }
                         // step 3 : add result to target bag.
-                        target.add( temp );
+                        target.ownProperty( temp );
                     }
             }
             ++it;
--- a/rtt/marsh/CPFDemarshaller.cpp
+++ b/rtt/marsh/CPFDemarshaller.cpp
@@ -143,10 +143,10 @@ namespace RTT
                         if ( type == "boolean" )
                         {
                             if ( value_string == "1" || value_string == "true")
-                                bag_stack.top().first->add
+                                bag_stack.top().first->ownProperty
                                 ( new Property<bool>( name, description, true ) );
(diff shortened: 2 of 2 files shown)


meyerj:
It should be noted that an alternative patch would introduce additional `RTT::deletePropertyBag()` and `RTT::deleteProperties()` calls in the `PropertyLoader` implementation to do the cleanup. But I think that using a `PropertyBag`'s built-in memory management (using `PropertBag::ownProperty()`) is the best option.

achim-k:
Looks like we have the same kind of leaks here as well?
https://github.com/orocos-toolchain/rtt/blob/d496d004818e0f01e3f6d9aa070af6062de4c40c/rtt/marsh/TinyDemarshaller.cpp#L99-L100

meyerj:
> Looks like we have the same kind of leaks here as well?
> https://github.com/orocos-toolchain/rtt/blob/d496d004818e0f01e3f6d9aa070af6062de4c40c/rtt/marsh/TinyDemarshaller.cpp#L99-L100

Indeed. Should use [deletePropertyBag()](https://github.com/orocos-toolchain/rtt/blob/54212b874d89a49836cb114a830ff3c4cf1e0fad/rtt/PropertyBag.hpp#L556) [here](https://github.com/orocos-toolchain/rtt/blob/d496d004818e0f01e3f6d9aa070af6062de4c40c/rtt/marsh/TinyDemarshaller.cpp#L434) and in all callers of `TinyDemarshaller::deserialize()`, or add properties to the bags using [ownProperty()](https://github.com/orocos-toolchain/rtt/blob/54212b874d89a49836cb114a830ff3c4cf1e0fad/rtt/PropertyBag.hpp#L190) instead.

PR #209, opened 2017-03-09
rtt Logger: clear any fail and eof flags

fengzhuye:
just call clear() will not reset the stringstream. And the log will not append to the stringsteam before reset the flag.
reflink: http://stackoverflow.com/questions/2848087/how-to-clear-stringstream


meyerj:
Looks good to me. Thanks for the contribution!

PR #211, opened 2017-04-03
Fix use after free during disconnection

doudou:


meyerj:
Why do you think the previous version was not correct?

`ChannelDescriptor descriptor` is a copy of the full Boost tuple holding the intrusive pointer to the `ChannelElement`. The copy assignment in [line 147](https://github.com/orocos-toolchain/rtt/pull/211/files#diff-04f28856860f9a8a2d4afc939c84e682R147) also increments the use count (and copies the `ConnPolicy` that contains a `std::string` element by the way) and prevents that the channel is destroyed before the end of the function. With your patch we could eventually avoid the copy and degrade `descriptor` to type `ChannelDescriptor *` or an iterator type, but in any case the `updateCurrentChannel()` method needs to be called before the use count reaches zero and the channel might be destroyed.

doudou:
> Why do you think the previous version was not correct?

Actually, valgrind does ;-)

`cur_channel` is a pointer to an element of `connections`, which may have been deleted by `erase`, causing `cur_channel->get<1>()` to do a use-after-free.

meyerj:
Ah yes, you are right. I was wrongly assuming that `cur_channel` is a pointer to `ChannelElementBase` only, but it is actually a pointer to the deleted descriptor.

A minor comment is that the variable should be renamed to `reset_current` to follow the naming conventions of other local and member variables in RTT and match the local name inside `updateCurrentChannel(bool reset_current)`. Otherwise the patch looks good to me.

Actually it would be much less confusing to simply set `cur_channel` to `NULL` if it matches `&(*conn_it)` and drop `updateCurrentChannel()` completely. Should have the same effect.

doudou:
> Actually it would be much less confusing to simply set cur_channel to NULL if it matches &(*conn_it) and drop updateCurrentChannel() completely. Should have the same effect.

I wouldn't mind doing that ... But then, getCurrentChannel's documentation assumes that as soon as there's a connection, there should be a current channel and addChannel enforces that.

meyerj:
> I wouldn't mind doing that ... But then, getCurrentChannel's documentation assumes that as soon as there's a connection, there should be a current channel and addChannel enforces that.

Not sure where else `getCurrentChannel()` is used, but at least the proposed patch would have an effect on [InputPort\<T\>::getDataSample()](https://github.com/orocos-toolchain/rtt/blob/9c46c957bcd6078995f4d36ebeec4da385ed8218/rtt/InputPort.hpp#L188) in some corner cases. So better keep it as it is now instead of updating the documentation.

doudou:
> So better keep it as it is now

OK. I'll change the style.

doudou:
Done. I've also added a comment in the code given how non-obvious this is ...

doudou:
Ping @meyerj

PR #215, opened 2017-05-16
TimerThread: do not return the same instance if CPU affinity differs

meyerj:
Unfortunately there is a regression bug in https://github.com/orocos-toolchain/rtt/pull/205 which has been merged into master.

The CPU affinity comparison in `RTT::TimerThread::Instance()` added in 2cfecd7 was not taking into account which CPUs are actually available and therefore created a new timer thread in almost all cases on target gnulinux. This patch makes `getCpuAffinity()` also available for `RTT::os::MainThread` and uses the main thread's CPU affinity as a default value for TimerThread instances.

If no specific CPU affinity was given in the Thread or Activity constructor, the `cpu_affinity` argument has a default value of 0. The gnulinux implementation of `rtos_task_create()` forwarded this value to `rtos_task_set_cpu_affinity()` which resets the affinity attributes to all CPUs in this case. On the other hand,  if `cpu_affinity` set by the user is `~0`, the call to rtos_task_set_cpu_affinity() was skipped and the new threads inherits the affinity of its creator, the default behavior of `pthread_create()`.

This patch basically swaps these two cases, so that new threads inherit the affinity of their parent (typically the main thread) and the affinity is only overwritten in case the user sets it explicitly. If the `cpu_affinity` argument is `~0`, the affinity will be reset to all available CPUs, disregarding the flags of the parent.

This might even be considered as a bug fix, looking at the test case in [tasks_test.cpp:376](https://github.com/meyerj/rtt/blob/f65d958ea07691525d0dd48ef6f3366bde9612a6/tests/tasks_test.cpp#L376).

@snrkiwi

### DIFF
--- a/rtt/os/xenomai/fosi_internal.cpp
+++ b/rtt/os/xenomai/fosi_internal.cpp
@@ -424,7 +424,7 @@ namespace RTT
 
         INTERNAL_QUAL unsigned rtos_task_get_cpu_affinity(const RTOS_TASK *task)
         {
-            return 0;
+            return ~0;
         }
 
         INTERNAL_QUAL const char* rtos_task_get_name(const RTOS_TASK* mytask) {
--- a/rtt/extras/TimerThread.cpp
+++ b/rtt/extras/TimerThread.cpp
@@ -44,6 +44,7 @@
 #include "../Logger.hpp"
 #include <algorithm>
 #include "../os/MutexLock.hpp"
+#include "../os/MainThread.hpp"
 
 namespace RTT {
     using namespace extras;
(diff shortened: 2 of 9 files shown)


rtt/os/xenomai/fosi_internal.cpp
```diff
@@ -424,7 +424,7 @@ namespace RTT
 
         INTERNAL_QUAL unsigned rtos_task_get_cpu_affinity(const RTOS_TASK *task)
         {
-            return 0;
+            return ~0;
```
meyerj: All other targets not supporting CPU affinity return `~0` here.

PR #216, opened 2017-05-18
Fix/create properly OrbRunner activity

MagnaboscoL:
In [TaskContextServer.cpp#L342](https://github.com/orocos-toolchain/rtt/blob/toolchain-2.9/rtt/transports/corba/TaskContextServer.cpp#L342) the `cpu_affinity` is passed as third argument and this way the compiler chooses to use the constructor in [Activity.hpp#L141](https://github.com/orocos-toolchain/rtt/blob/toolchain-2.9/rtt/Activity.hpp#L141) where the third argument is the period (double).

This PR proposes also a naming convention for the Threads created by RTT itself.
For the moment the new naming convention is used only for the OrbRunner Thread.

@meyerj PTAL

PS
If preferred I could rebase the PR on master.

### DIFF
--- a/rtt/transports/corba/TaskContextServer.cpp
+++ b/rtt/transports/corba/TaskContextServer.cpp
@@ -339,7 +339,7 @@ namespace RTT
     {
     public:
         OrbRunner(int scheduler, int priority, unsigned cpu_affinity)
-            : Activity(scheduler, priority, cpu_affinity)
+            : Activity(scheduler, priority, 0.0, cpu_affinity, 0, "OrbRunner")
         {}
         void loop()
         {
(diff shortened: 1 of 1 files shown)


meyerj:
Looks good to me. Thanks!

I am not aware of a written down policy regarding commit messages (@psoetens?), but it is common to [sign off](https://github.com/git/git/blob/b2c150d3aa82f6583b9aadfecc5f8fa1c74aca09/Documentation/git-commit.txt#L154-L161) commits within RTT and OCL (using `git commit -s`). I also typically prefix commit messages with a short keyword or path, e.g. `engine:`, `os:` or `corba:` in this case, but that is probably a matter of taste. Ideally the commit message also contains the full explanation/motivation of the patch if it is not obvious, because it is much easier to search and also available in the local history without having to depend on a specific Git service provider like GitHub. I usually copy the exact same text to the pull request afterwards.

MagnaboscoL:
This PR is closed since Internal RTT Threads used by OrbRunner and CorbaDispatcher should also be parametrized as proposed in [PR218](https://github.com/orocos-toolchain/rtt/pull/218).

PR #217, opened 2017-05-19
rtt: fix null ptr exception in RTT::TaskContext::setActivity

MagnaboscoL:
Fix null pointer exception that was generated only if the following conditions were met:
- `ORO_ACT_DEFAULT_SEQUENTIAL` not defined
- `ORO_ACT_DEFAULT_ACTIVITY` not defined
- `new_act == 0`

@meyerj PTAL

### DIFF
--- a/rtt/TaskContext.cpp
+++ b/rtt/TaskContext.cpp
@@ -344,22 +344,25 @@ namespace RTT
     {
         if (this->isRunning())
             return false;
-        if ( new_act == 0) {
+        if (!new_act) {
 #if defined(ORO_ACT_DEFAULT_SEQUENTIAL)
             new_act = new SequentialActivity();
 #elif defined(ORO_ACT_DEFAULT_ACTIVITY)
             new_act = new Activity();
-#else
-            return false;
 #endif
+        } else {
+            new_act->stop();
         }
-        new_act->stop();
-        if(our_act){
+        if (our_act) {
             our_act->stop();
         }
-        new_act->run( this->engine() );
-        our_act = ActivityInterface::shared_ptr( new_act );
-        our_act->start();
+        if (new_act) {
+            new_act->run( this->engine() );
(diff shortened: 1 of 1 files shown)


rtt/TaskContext.cpp
```diff
@@ -349,6 +349,8 @@ namespace RTT
             new_act = new SequentialActivity();
 #elif defined(ORO_ACT_DEFAULT_ACTIVITY)
             new_act = new Activity();
+#else
+            return false;
```
meyerj: This way you cannot clear the activity of a TaskContext anymore. Not sure if this use case would ever be required and it was broken anyway, but as having no activity (`our_act == NULL`) is the default state after construction if neither `ORO_ACT_DEFAULT_SEQUENTIAL` or `ORO_ACT_DEFAULT_ACTIVITY` is defined there is no reason why it should not be possible to also clear the activity using `setActivity(NULL)`. So instead of returning false here you could also check whether `new_act` is set in the remainder of the method and if not, only stop and destroy `our_act` if set.

MagnaboscoL: Not sure if I understand what you mean.
If `new_act == 0` but `ORO_ACT_DEFAULT_SEQUENTIAL` or  `ORO_ACT_DEFAULT_ACTIVITY` are defined  a new default activity of the proper type is created and the old one (if existing) is stopped.
So wouldn't be inconsistent to use this method to clear the activity of the TaskContext only if both definitions are not set?

meyerj: If no specific default activity type is given in the RTT configuration (neither `ORO_ACT_DEFAULT_SEQUENTIAL` nor `ORO_ACT_DEFAULT_ACTIVITY` is defined), the TaskContext could have no activity attached. I think this has been the default some time ago and is still a valid option.

PR #218, opened 2017-05-19
transports::corba:: add configuration Properties for OrbRunner and CorbaDispatcher Threads

MagnaboscoL:
This PR proposes a possible approach to make the scheduler, the priority and the CPU affinity of the RTT internal OrbRunner and CorbaDispatcher Threads configurable.

**About CorbaDispatcher Thread configuration**
The proposed approach is based on the fact that only a  CorbaDispatcher per Component is created while the ConnPolicy, that typically control the "properties" of a Port, can be different for every couple of connected ports.
The Properties `CorbaDispatcherScheduler`,  `CorbaDispatcherPriority` and `CorbaDispatcherCpuAffinity` should be set before connecting the first Corba port. 

**About OrbRunner Thread configuration**
Since the OrbRunner Thread is created before parsing the normal xml files the `site-file` should be used to set the Properties`OrbRunnerScheduler`, `OrbRunnerPriority` and `OrbRunnerCpuAffinity`.

An example of site-file usage is:
`deploy deployer-corba-$OROCOS_TARGET --site-file <path>/site-file_name.xml -s <path>/conf_file_1_name.xml  -s <path>/conf_file_2_name.xml`

A site-file example is:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE properties SYSTEM "cpf.dtd">
<properties>
  <!-- ocl needs to be imported explicitly if the site-file is specified -->
  <simple name="Import" type="string"><value>ocl</value></simple>
  <struct name="GlobalsRepository" type="">
    <struct name="Properties" type="PropertyBag">
      <!-- OrbRunner Properties have to be loaded with the site-file -->
      <simple name="OrbRunnerScheduler" type="long"><value>1</value></simple>
      <simple name="OrbRunnerPriority" type="long"><value>1</value></simple>
      <simple name="OrbRunnerCpuAffinity" type="long"><value>1</value></simple>
    </struct>
  </struct>
</properties>
```

To load the Properties described above **ocl** should be able to handle the `GlobalsRepository` tag, so the modifications proposed by [PR63](https://github.com/orocos-toolchain/ocl/pull/63) are needed.

PS
To guarantee that the Properties are loaded before the ports are connected (even if AutoConnect is set) this commit could be used [link](https://github.com/MagnaboscoL/ocl/commit/ad32834c67a423fcdce6820b589365c4eddeaaea).

@meyerj PTAL


snrkiwi:
How would this be used? How would a project using the OCL deployer + XML set these properties? Or using the Deployer + a script? Are you envisioning that these properties are settable per component?

MagnaboscoL:
These Properties can be indeed set "per Component" and this is useful because the "CorbaDispatcher Thread" is instantiated "per Component" too.

The code has been tested loading these properties from an XML where the Components to be loaded by the Deployer are specified.
To set these Properties is enough to add to the XML the following lines inside the "xml struct" that load the specific Component
```
<struct name="MyComponentInstanceName" type="MyComponent">
   ...
   ...
    <struct name="Properties" type="PropertyBag">
	<simple name="CorbaDispatcherScheduler" type="long"><value>1</value></simple>
	<simple name="CorbaDispatcherPriority" type="long"><value>1</value></simple>
    </struct>
   ...
   ...
</struct>
```

snrkiwi:
Looks quite reasonable. Any reason that the user can't also set the scheduler and affinity?

I'm also wondering whether its worth considering a global default set of parameters as well, which is used if the user doesn't specify parameters per component? This would also allow complex applications that just want to shove all the network traffic on say one (1) CPU to only have to set one set of parameters, rather than a set of parameters for each of N components.

MagnaboscoL:
> Any reason that the user can't also set the scheduler and affinity?

The `scheduler`could be set with the proposed solution, are you referring to another scheduler?
As for the `affinity`, it could be surely set, the modification would just require to add the argument in the Dispatcher constructor and to use the proper constructor for the underlying Activity.

> I'm also wondering whether its worth considering a global default set of parameters as well, which is used if the user doesn't specify parameters per component? This would also allow complex applications that just want to shove all the network traffic on say one (1) CPU to only have to set one set of parameters, rather than a set of parameters for each of N components.

If I am not wrong, this could be easily achieved using as "default set of parameters" some Properties with the same name assigned to the Deployer. To achieve that, a minor modification to the proposed code should be implemented:
1. if property is set for the Component then use it. If not go to next rule. [Already implemented]
2. if property is set for the `Deployer` (a peer of our Component that is normally available "per application") then use it. If not go to next rule. [To be implemented]
3. use default hard coded value [Already implemented]

My main doubt is still about using `Properties` or `Constants`. 
The usage of Properties probably isn't the best solution because the value of a Property can be updated multiple times while these parameters could be changed only before the "CorbaDispatcher Thread" is created. Can we assume that once set (per Component) these parameters can't be changed? 
If yes the usage of Constants would probably be a better fit. 
If it is the case, a small extra work is needed because constants are not loaded from the XML yet.

MagnaboscoL:
Last commit switch to use `GlobalsRepository` Properties instead of `GlobalService` since the second one is intended more for services while the first one is already used for some global Properties.

rtt/transports/corba/CorbaDispatcher.hpp
```diff
@@ -111,7 +113,30 @@ namespace RTT {
                     else
                         name = iface->getOwner()->getName();
                     name += ".CorbaDispatch";
-                    DispatchI[iface] = new CorbaDispatcher( name, scheduler, priority );
+
+                    // The properties to create the CorbaDispatcher are retrieved.
+                    // When the CorbaDispatcher is created these properties can't be changed anymore,
+                    // so they are converted to Constants.
+                    TaskContext* owner = iface->getOwner();
```
meyerj: `owner` might be `NULL`

meyerj: Fixed in https://github.com/orocos-toolchain/rtt/pull/218/commits/285354c5792f8a25c30e4f6b171da531e647ebae.

PR #219, opened 2017-05-29
add aarch64 support

ahoarau:
This just add support for the aarch64 tag (arm64).
Sorry for the patch-1 name, I did it with the web interface.

### DIFF
--- a/rtt/rtt-config.h.in
+++ b/rtt/rtt-config.h.in
@@ -35,6 +35,8 @@
 #  define OROBLD_OS_ARCH_ppc
 # elif defined( __GNUC__ ) && (defined( __arm__ ) || defined( __ARM__ ) )
 #  define OROBLD_OS_ARCH_arm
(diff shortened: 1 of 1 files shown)


meyerj:
I could not find any usage of these macros within RTT except for [rtt/os/oro_arch.h](https://github.com/ahoarau/rtt/blob/d1efde9efee66f8db418674688ca3df9221a3b97/rtt/os/oro_arch.h), but there is no special case for the ARM architecture anyway. What is the effect of the added tag? Only for derived packages?

meyerj:
Build failure on Travis would be fixed by #215.

ahoarau:
I don't use it, but it's necessary otherwise it send an error "unknown processor architecture".

ahoarau:
Thanks for this ! Does it mean rtt (+everything else) will be available in the arm64v8 build farm ?

ahoarau:
@meyerj aby way to trigger the debian build for aarch64 (arm64v8) ?

PR #221, opened 2017-06-08
rm some unnecessary Logger::In

fengzhuye:
remove some unneccessary Logger::In print in StateMachine & OperationCallerC.

### DIFF
--- a/rtt/scripting/StateMachine.cpp
+++ b/rtt/scripting/StateMachine.cpp
@@ -47,9 +47,12 @@
 #include <boost/tuple/tuple.hpp>
 #include "internal/mystd.hpp"
 
-//#define TRACE_INIT() Logger::In in( _name )
-#define TRACE_INIT() 
-#define TRACE(msg) if (mtrace) log(Info) << '[' << this->getStatusStr() << ']' << std::string(" ") + msg <<endlog()
+#define TRACE(msg) do {                                                                     \
+    if (!mtrace)                                                                            \
+      break;                                                                                \
+    Logger::In in( _name );                                                                 \
+    log(Info) << '[' << this->getStatusStr() << ']' << std::string(" ") + msg <<endlog();   \
+} while(0)
 
 namespace RTT {
     using namespace detail;
@@ -71,7 +74,6 @@ namespace RTT {
 
    StateMachine::~StateMachine()
     {
-       TRACE_INIT();
        if ( this->isLoaded() ){
            getEngine()->removeFunction(this);
        }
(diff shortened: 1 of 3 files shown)


rtt/scripting/StateMachine.cpp
```diff
@@ -47,7 +47,7 @@
 #include <boost/tuple/tuple.hpp>
 #include "internal/mystd.hpp"
 
-#define TRACE_INIT() Logger::In in( _name )
+#define TRACE_INIT() if (mtrace) Logger::In in( _name )
```
meyerj: I agree that some `Logger::In` might be unnecessary and cause a small overhead, but this one is actually wrong: `Logger::In in` must remain valid for the remainder of the function calling `TRACE_INIT()`. With your proposed patch it would be a no-op because `in` immediately goes out of scope again after that line.

I think you would need a patch for the `RTT::Logger::In` class to allow for passing a condition to the constructor. Should be possible in an ABI- and API-compatible way.

meyerj:
Thanks for the contribution. Some details about your motivation would be nice. Are those lines breaking your application or cause real-timeliness issues?

Also note that `RTT::Logger::In` is actually broken since a long time as it is not thread safe, which is definitely a bug for a real-time toolkit. The module name member of each `Logger` instance would have to be stored per thread. Otherwise a thread A that restores the module name after one section might change it right after a thread B set it and before B wants to print a log line, which is a classical race condition.

fengzhuye:
Indeede, there are some thread-safety problem related to the Logger::In. For example, there is a change than one will access the Logger::In from xenomai thread in Primary Mode and Secondary Mode, then the mutex inside the Logger::In cannot protect the resources.

This PR is about to rm the Logger::In in StateMachine loop, because it will cause the log modual name messy. Besides, it add some overhead. 

You are right, this patch is useless, maybe it will be better if i just disable all the Logger::In inside the StateMachine like this:
  `#define TRACE_INIT() Logger::In in( _name )` ? @meyerj

fengzhuye:
I find there are some problem with stringstream's FIFO implementation. The memory of the stringstream will not be freed until the stringstream object is distroyed. 

This latest commit of this PR replace the stringstream with std::deque<std::string>. The reason not using boost::circular_buffer is that i found the Logger.cpp not including any boost header.

meyerj:
I really like the cleanup of state machine tracing and the replacement of the `stringstream` by `BufferLockFree<std::string>`. Thanks for your contribution!

One additional question: Are you actually using remote logging or was the patch necessary as a consequence of the other changes?

I wonder whether we should disable `OROSEM_REMOTE_LOGGING` by default. It has not been touched since a long time before and I do not see a valid use case. I personally was not aware that all log messages are pushed to a `stringstream` buffer that probably nobody ever really uses.

fengzhuye:
@meyerj Yes, I actually using the remote logging in our project. 

As we want to display the logs on the UI which is connected to the orocos component through network.

PR #223, opened 2017-08-23
types: fix type lookup for qualified types

MagnaboscoL:
This PR fixes type lookup for qualified types.

Caveat:
Types with spaces in their name will not be supported anymore.

@meyerj PTAL


meyerj:
This patch was originally written to fix a bug related to CORBA operation calls introduced in #123. I added a commit there which is less intrusive and only affects the CORBA transport layer instead of all type lookups: https://github.com/orocos-toolchain/rtt/pull/123/commits/a364a3759ef7cc9e863053ce6cd667feea125dc2.

PR #224, opened 2017-09-22
Xenomai 3 support

ahoarau:
This PR allows to build and run applications on Xenomai 2 and 3 with respectively the native and alchemy skins.

#### Why alchemy rather than posix

I choose to use the alchemy (formerly 'native') interface to avoid all the magic xenomai function wrappers that replaces standard posix calls to xenomai posix calls under the hood. It's way clearer to use alchemy in the end.

#### Xenomai 3 installation guidelines

http://rtt-lwr.readthedocs.io/en/latest/rtpc/xenomai3.html

#### Open issues

* TLSF says it has an overhead (alloc - free > 0) (also on gnulinux build)
* The tlsf-test.cpp fails for some reason, on both xeno 2 and 3. TLSF overhead might be the issue.
* There are 2 context switches when creating a task, that points to the mallocs used in rtos_task_create
* In the current xenomai 2 implementation, ```ORO_SCHED_OTHER``` does **not** force the task priority to zero. In xeno 3, prio zero means ```SCHED_OTHER```, it's unclear to me for xeno 2. (ref : https://xenomai.org/documentation/xenomai-3/html/xeno3prm/group__alchemy__task.html#ga03387550693c21d0223f739570ccd992)
* Copperplate uses a `debug()` macro in C, which conflicts with log4cpp. Only hack I found was to add `#define _COPPERPLATE_DEBUG_H` in `fosi.h`. Not very elegant. 
* Is there any way to export the CMake modules present in config so other packages can use them without copy/pasting them ?

#### Test systems

* Ubuntu 16.04 - kernel 4.9.38   - xenomai 3.0.5
* Ubuntu 14.04 - kernel 3.18.20 - xenomai 2.6.5

OROCOS + rtt_ros_integration compiled with :
```
export OROCOS_TARGET=xenomai
catkin config --cmake-args -DCMAKE_BUILD_TYPE=Release -DENABLE_MQ=ON -DENABLE_CORBA=ON -DCORBA_IMPLEMENTATION=OMNIORB -DENABLE_TESTS=ON
```

Thanks in advance for taking the time to review this. 
PS: You can squash those commits instead of merging to you won't have all the debugging


meyerj:
Thanks, @ahoarau. Patches to revive and upgrade Xenomai support are very welcome, but I personally or Intermodalics have not used the Xenomai target anymore for several years besides testing patches and running the unit tests. I hope I will find some time to review it as a whole in the near future. If you have specific questions, I can probably quickly comment on it.

> @meyerj I'd need your guidance on this one. Surprisingly the mutex wasn't always lock before waiting on the condition. xeno2 seems to be fine with it. not xeno3.

Which mutex exactly are you referring to?

ahoarau:
> Patches to revive and upgrade Xenomai support are very welcome, but I personally or Intermodalics have not used the Xenomai target anymore for several years besides testing patches and running the unit tests.

Interesting. Do you use `PREEMPT_RT` patches instead ? I remember having too much jitter to control  the Kuka LWR at 1Khz, but I'm guessing on production systems you don't have 10 chromes tabs openned + atom + kdevelop that can distract the kernel from sending commands.

> If you have specific questions, I can probably quickly comment on it.

The "open issues" section is a good starting point :) 

I can add another question regarding the `FindXenomai.cmake` and `FindXenomaiPosix.cmake`. I've basically updated the existing ones to support xeno3, but we could also only provide one `FindXenomai.cmake` that can provide `XENOMAI_LIBRAIRIES` (native/alchemy skin), `XENOMAI_RTDM_LIBRAIRIES` (rtdm support) and `XENOMAI_POSIX_LIBRAIRIES` (posix skin).
An example can be found here : https://github.com/kuka-isir/ati_sensor/blob/master/config/FindXenomai.cmake

```
find_package(Xenomai)
# provides _LIBRAIRIES, INCLUDE_DIRS, CFLAGS, LD_FLAGS, COMPILE_DEFINITIONS for 
# XENOMAI_
# XENOMAI_POSIX_
# XENOMAI_RTDM_
```

> Which mutex exactly are you referring to?

The changes from this commit : 3345b98

ahoarau:
xenomai 3 does not send any error when broadcasting on unlocked mutex (returns 0). The one waiting on the condition is waiting forever.

snrkiwi:
Some comments
- there are a bunch of whitespace-only changes unrelated to Xenomai in `config/check_depend.cmake, tests/CMakeLists.txt` (at least). These can make merging and/or diffs complicated later.
- are there a bunch of changes to global CMake var's in `config/check_depend.cmake` that are no longer wrapped by `if(XENOMAI_FOUND)`?
- yeah, those changes in `rtt/Activity.cpp, rtt/ExecutionEngine.cpp, ...` scare me. The side effects on other O/S's could be interesting ... If we're not sure of their affect then I would suggest conditionally making those changes only in Xenomai.
- the changes in `rtt/os/tlsf/tlsf.c` are not backwards compatible, and will affect other people's installations. I think that changing *your* project to set the default `TLSF_USE_LOCKS` value in  the `rtt.default.cmake` (sp?) would be more suitable.
- should the `testDoubleFree` in `tests/tlsf_test.cpp` only be "commented out" on Xenomai?

Note that TLSF should have an overhead, as it takes its meta data out of the pool of memory that you give it. This is a few kb IIRC>

ahoarau:
> there are a bunch of whitespace-only changes unrelated to Xenomai in config/check_depend.cmake, tests/CMakeLists.txt (at least). These can make merging and/or diffs complicated later.

I'll try to make a properer PR soon.

> are there a bunch of changes to global CMake var's in config/check_depend.cmake that are no longer wrapped by if(XENOMAI_FOUND)?

Just above you have `find_package(Xenomai REQUIRED)`, so I guess you're supposed to find it, otherwise it will send and error.

> yeah, those changes in rtt/Activity.cpp, rtt/ExecutionEngine.cpp, ... scare me. The side effects on other O/S's could be interesting ... If we're not sure of their affect then I would suggest conditionally making those changes only in Xenomai.

I'll wrap them around some xenomai compile flag.

> the changes in rtt/os/tlsf/tlsf.c are not backwards compatible, and will affect other people's installations. I think that changing your project to set the default TLSF_USE_LOCKS value in the rtt.default.cmake (sp?) would be more suitable.
> should the testDoubleFree in tests/tlsf_test.cpp only be "commented out" on Xenomai?

I'll run some extra tests now that everything is working properly.

PR #228, opened 2017-10-16
Allow loading of global service plugins via the CORBA interface of each TaskContextProxy

meyerj:
Follow-up of #134, which reverted a patch introduced in #90 for loading global service plugins into an Orocos process from remote peers, which is required for the RTT introspection feature. As far as I know this use case has not yet been solved or worked around differently. @jmachowinski?

From https://github.com/orocos-toolchain/rtt/pull/134#issuecomment-280730858:
> The second approach is probably more feasible and similar to the operation you added to the TaskContext interface, but I would only expose the access to PluginLoader::loadPlugin() (or all global services) in TaskContextProxy instances and not for every local TaskContext. And it would not break existing code that already uses the loadPlugin(...) operation on the remote side.

From https://github.com/orocos-toolchain/rtt/pull/134#issuecomment-280748429:
> I created a patch that implements the second solution: [master...meyerj:fix/load-plugin-for-introspection](https://github.com/orocos-toolchain/rtt/compare/master...meyerj:fix/load-plugin-for-introspection).
> As promised, it is a bit hacky, but would solve the problem. Feel free to create a new PR if the patch also works for you, but it is more a proof-of-concept than something I would really consider to merge. Probably it would be better to implement a PluginLoaderService, add it to the GlobalService and expose this global service to every TaskContextProxy as a fake service "Global" instead of adding only one fake operation.

### DIFF
--- a/rtt/internal/FusedFunctorDataSource.hpp
+++ b/rtt/internal/FusedFunctorDataSource.hpp
@@ -45,6 +45,7 @@
 #include "BindStorage.hpp"
 #include "../ExecutionEngine.hpp"
 #include "../os/oro_allocator.hpp"
+#include "UnMember.hpp"
 #include <boost/bind.hpp>
 #include <boost/type_traits.hpp>
 #include <boost/function.hpp>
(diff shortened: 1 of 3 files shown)


jmachowinski:
I'm using your patch, and it is working. Still it is kind of ugly and hacky, but as long as it works its fine for me.

jmachowinski:
Ohh, I just read the part about the merge request I was supposed to create, sorry, forgot / overread that part...

PR #230, opened 2017-10-17
Corba aliases: revert API changes and fixed ComponentLoader::unloadComponent() for aliased components

meyerj:
#66 introduced aliases for CORBA component servers and added an optional name argument to `ComponentLoader::unloadComponent(tc)` in bf92dcb3483f89ccb436502e1f74a04361e53c74. This change has been reverted, because it could lead to inconsistencies in the internal component map, if a TaskContext is destroyed based on the given name but with an unrelated instance pointer.

This concern has initially be brought up as a comment in https://github.com/orocos-toolchain/rtt/commit/bf92dcb3483f89ccb436502e1f74a04361e53c74.

The new approach should be safe as it does not rely on the name or alias of the instance to be unloaded at all.


snrkiwi:
Looks good

PR #232, opened 2017-10-22
Fixed bug in component unloading.

erniejunior:
This while loop never terminates if there is more than one component loaded. Ouch!


meyerj:
Oops, thanks for catching.
This is a regression from #230 that has been merged 8 days ago.

PR #237, opened 2017-10-26
use another variable name for CMAKE_PREFIX_PATH entries

### LINKED ISSUE #225: Sourcing ROS with rtt screws up zsh environment
Hi there, 

Yesterday I realized, that my zsh-env got screwed up as soon as the ros-kinetic-rtt package was installed:

```
> $ source /opt/ros/kinetic/setup.zsh                                                                                                                                                       
/opt/ros/kinetic/etc/catkin/profile.d/00.rtt.sh:10: command not found: grep
/opt/ros/kinetic/etc/catkin/profile.d/10.ros.sh:12: command not found: cat
/opt/ros/kinetic/etc/catkin/profile.d/99.roslisp.sh:4: command not found: cat
/opt/ros/kinetic/etc/catkin/profile.d/99.roslisp.sh:21: command not found: python
```
I nailed it down to the for loop of the file 00.rtt.sh 
`for path in `echo $CMAKE_PREFIX_PATH | tr : ' '` "@CMAKE_INSTALL_PREFIX@"; do`

Apparently `path` is reserved (or even equivalent) to `PATH`, so this for loop overwrites the PATH. Renaming the variable into anything else will solve the problem.

fmauch:
in a zsh environment $path is the same as $PATH. So having this package
installed and sourcing ROS screws up everything.

Fixes #225

### DIFF
--- a/env-hooks/00.rtt.sh.in
+++ b/env-hooks/00.rtt.sh.in
@@ -2,13 +2,13 @@
 
 : ${OROCOS_TARGET:=@OROCOS_TARGET@}
 
-for path in `echo $CMAKE_PREFIX_PATH | tr : ' '` "@CMAKE_INSTALL_PREFIX@"; do
-  if [ ! -d "$path/lib/orocos" ]; then
+for file_path in `echo $CMAKE_PREFIX_PATH | tr : ' '` "@CMAKE_INSTALL_PREFIX@"; do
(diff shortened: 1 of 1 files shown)


meyerj:
Thanks for the patch, looks good to me.

We should probably also apply this patch to the toolchain-2.8 and toolchain-2.9 branches and re-release into ROS asap.

PR #239, opened 2017-10-27
Fix/add TypeInfo retrieval by alias name

meyerj:
The type name aliases registered in the TypeInfo are currently unused within RTT and OCL. This commit allows to lookup TypeInfo pointers via an alias, e.g. in Orocos scripting. Not sure what they have been used for before.

Cherry-picked, slightly modified and reopened from https://github.com/meyerj/rtt/pull/5.


meyerj:
Sorry, this was a duplicate of https://github.com/orocos-toolchain/rtt/pull/223. I was not aware that there was already a PR to master when I created this one.

PR #240, opened 2017-10-31
fix compilation on win32 as an orogen target

doudou:
This allows to compile an orogen-generated package on win32.

The one change that I really am not sure about is https://github.com/orocos-toolchain/rtt/commit/55835714594bf28c30fdad032939fb3fe4317571. Since I have very little knowledge of TAO, I am afraid to break something on other TAO versions or targets. @meyerj, an idea ?


meyerj:
Looks good to me.

> The one change that I really am not sure about is 5583571. Since I have very little knowledge of TAO, I am afraid to break something on other TAO versions or targets. @meyerj, an idea ?

A related commit to fix builds with TAO as a CORBA implementation, https://github.com/orocos-toolchain/rtt/commit/bc6503b7012f42a4d08a062839dee5fdead02ed5, has been merged into `toolchain-2.9`. I am also not familiar with TAO and simply assume that newer versions do not generate that file anymore. Should be fine if we both came to the same solution.

PR #243, opened 2017-11-13
Install headers to include/orocos/rtt instead of include/rtt

meyerj:
...for consistency with [log4cpp](https://github.com/orocos-toolchain/log4cpp/blob/toolchain-2.9/CMakeLists.txt#L135) (since toolchain-2.9), [ocl](https://github.com/orocos-toolchain/ocl/blob/toolchain-2.9/ocl/CMakeLists.txt#L24) and the default include directory for Orocos packages (see [orocos_install_headers()](https://github.com/meyerj/rtt/blob/f95735acc7ddc7509f0ac556f57212812ae1a798/UseOROCOS-RTT.cmake#L810) and [orocos_generate_package()](https://github.com/meyerj/rtt/blob/f95735acc7ddc7509f0ac556f57212812ae1a798/UseOROCOS-RTT.cmake#L937) macros in UseOrocos-RTT.cmake).

The renaming has primarily practical reasons during RTT development, if one version of RTT is installed (e.g. from Debian package) while working on another, incompatible version in a custom workspace. If RTT installs its headers directly to `{prefix}/include/rtt` (e.g. `/usr/include/rtt`) it cannot be avoided that dependents find the installed headers instead, which is likely to cause linker errors or segmentation faults.

This commit solves the problem at least for the case that RTT and all its dependents are present in the overlayed workspace and `{prefix}/include/orocos` of the installed version is not in the header search path.

Any objectives?

TODO:
- [rtt_typelib](https://github.com/orocos-toolchain/rtt_typelib) installs headers to `include/rtt/typelib`, which should also be changed to `include/orocos/rtt/typelib` then (see [rtt_typelib/CMakeLists.txt:46](https://github.com/orocos-toolchain/rtt_typelib/blob/toolchain-2.9/CMakeLists.txt#L46)). All other Orocos packages I am currently aware of are already installing their headers to `include/orocos` or use the `orocos_install_headers()` macro without overriding the default destination directory.


meyerj:
As there were no objections since 2017, we are going to merge this into `toolchain-2.10` now, which will soon be merged into `master` as a whole.

Issue #235, opened 2017-10-24
RTT visibility support?


meyerj:
Background information: https://gcc.gnu.org/wiki/Visibility

A discussion related to #234 would be the state of hidden visibility support of Orocos RTT in general. By default RTT does not add the `-fvisibility=hidden` or `-fvisibility-inlines-hidden` compiler options, so unless this is done by the user, all symbols will be visible. The latter has been commented out in 7724f0db47b3de47659d8fd112d7aa200a5f2ae1 for some reason and never has been reactivated again since 2007.

I tried both options with Ubuntu Xenial and gcc 5.4.1 and the current [toolchain-2.9](https://github.com/orocos-toolchain/rtt/tree/ce3328a1970a47d17f0ad2ef0b37735a7c91caaf) versions of RTT and OCL. Adding `-fvisibility=hidden` results in linker errors for the typekit and transport plugins. But `-fvisibility-inlines-hidden` works and only slightly reduces the size of the binaries, but cuts the length of the dynamic symbol table of `liborocos-rtt-gnulinux.so` by half. After [stripping](https://linux.die.net/man/1/strip) unneeded symbols from both libraries the size of the binary compiled with `-fvisibility-inlines-hidden` was only half of the one produced by a default build. For the RTT typekit plugin library the result is basically the same.

There have been [quite some patches](https://github.com/orocos-toolchain/rtt/search?q=visibility&type=Issues) to the build config recently to set the `GCC_HAS_VISIBILITY` cmake flag and preprocessor macro for different compilers and actually it is probably not used at all (with the exception of some sparse explicit `RTT_HIDE` declarations) and a potential source of errors when creating patches without taking care of `RTT_API` declarations.

Is there anyone who is compiling RTT with non-default build options or who has experience with hidden visibility builds from before 2009?

PR #245, opened 2017-11-30
Xenomai 3 support v2

ahoarau:
This is a follow-up to #224, with a proper PR this time.
Thanks for reviewing this !


meyerj:
I was testing this patch in a virtual machine in order to test if #258 would break the xenomai target. I needed some minor adaptations of the cmake code that decomposes the compiler and linker flags as returned by `xeno-config`. I could successfully compile with Xenomai 3.0.3, but the unit tests did freeze the virtual machine at some point. I had to upgrade the user-space libraries from 3.0.2 to 3.0.3 because `--auto-init-solib` was not yet available in 3.0.2, but the kernel patch was still 3.0.2. Not sure if this could explain the freeze or other odd behavior.

I would like to try again with a cleaner setup at some point, but I am also fine with merging as long as the patch does not break other targets.

Do you want me to push the cmake patch to your `xenomai3-support-v2` directly or do you prefer a separate PR?

rtt/Activity.cpp
```diff
@@ -163,7 +168,12 @@ namespace RTT
             return false;
         }
         mtimeout = true;
-        msg_cond.broadcast();
+        {
+#if defined(OROPKG_OS_XENOMAI) && (CONFIG_XENO_VERSION_MAJOR == 3)
+            os::MutexLock lock(msg_lock);
```
meyerj: Locking the mutex here is potentially critical because it creates a synchronization point between a thread that sends an operation and the one that has to process it. We had problems with priority inversions in the past even though the critical section is rather short.

Does Xenomai require to hold the mutex while broadcasting the condition variable?

rtt/internal/SharedConnection.cpp
```diff
@@ -63,16 +63,14 @@ ConnID* SharedConnID::clone() const
...
-static boost::uuids::random_generator uuid_generator;
-#endif
 
 SharedConnectionBase::SharedConnectionBase(const ConnPolicy &policy)
     : policy(policy)
 {
     // assign random name if none was given
     if (this->policy.name_id.empty()) {
 #if defined(ORO_HAVE_BOOST_UUID)
+        static boost::uuids::random_generator uuid_generator;
```
meyerj: :+1:

rtt/Activity.cpp
meyerj: Ah, this has already been answered in https://github.com/orocos-toolchain/rtt/pull/224#issuecomment-336903027.

meyerj:
I cannot really help with the open issues related to TLSF or Xenomai internals mentioned in #224, in case they are still present in this PR, because of a lack of time and recent experience with Xenomai. We run unit tests regularly for the `gnulinux` target and there are no known issues with the `tlsf_test`.

The RTT version of TLSF has some patches and it is very likely that the mismatch causes problems if the Xenomai libraries would provide the same symbol names. If that's the case, we would have to rename the functions in RTT or control visibility (https://github.com/orocos-toolchain/rtt/issues/235) in order to hide them from linked libraries or user code.

ahoarau:
> I was testing this patch in a virtual machine in order to test if #258 would break the xenomai target. I needed some minor adaptations of the cmake code that decomposes the compiler and linker flags as returned by xeno-config. I could successfully compile with Xenomai 3.0.3, but the unit tests did freeze the virtual machine at some point. I had to upgrade the user-space libraries from 3.0.2 to 3.0.3 because --auto-init-solib was not yet available in 3.0.2, but the kernel patch was still 3.0.2. Not sure if this could explain the freeze or other odd behavior.

I was using 3.0.5, and then the git master branch at some point to get the latests fixes. That's what Phillipe Gerum recommended to many people on the mailing list.

> I would like to try again with a cleaner setup at some point, but I am also fine with merging as long as the patch does not break other targets.

Tests I ran were surely not enough. The patched orocos version 2.9+Xenomai2 was tested ok at ISIR in october 2017, I made sure that the build flags were the same as before the patch. The patch is also running in a dev setup with xenomai3 and an iiwaa arm in another installation (to control it at 1KHz), since december 2017. So far so good.

> Do you want me to push the cmake patch to your xenomai3-support-v2 directly or do you prefer a separate PR?

Please push on this branch if it's any faster. Might be a good time though to bring the camelCase consistency in the FindXenomai cmake file, i.e. having `Xenomai_LIBRARIES` instead of `XENOMAI_LIBRARIES`. 
And maybe integrate RTnet directly to avoid having another file, and making the xeno2/3 support a bit less painful : https://github.com/kuka-isir/ati_sensor/blob/master/CMakeLists.txt#L30:L54

meyerj:
I suggested some updates in https://github.com/ahoarau/rtt/pull/1 that are needed for Xenomai 2.6.5, but I have not tested them recently.

ahoarau:
I dont have my xenomai setup anymore. So far the xeno3 I setup is still working flawlessly, but at an old commit. Are tests working again ?

meyerj:
> I dont have my xenomai setup anymore. So far the xeno3 I setup is still working flawlessly, but at an old commit. Are tests working again ?

I did not test with Xenomai 2.6 or 3.0 recently. Just opened the PR https://github.com/ahoarau/rtt/pull/1 to keep track of it.

In any case we should not merge this PR before it has been verified that it does not break Xenomai 2.6 and that the unit tests succeed.

PR #247, opened 2017-12-01
Fix setting the caller engine for RemoteOperationCaller instances

meyerj:
This bugfix affects all OperationCallers where the signature does not exactly match the operation's signature. In this case `RTT::OperationCaller<Signature>` falls back to an `RemoteOperationCaller<Signature>` implementation, which is able to adapt argument or return types if the exact types or qualifiers do not match.

Without this patch it was not possible to set the caller engine of such an operation caller after the initial construction because it did not reimplement the [OperationCallerInterface::setCaller()](https://orocos-toolchain.github.io/rtt/master/api/html/structRTT_1_1base_1_1OperationCallerInterface.html#adb68aecd6596ce7e6760dd5e3da999e2) method, but stored the original caller inside the `OperationCallerC` instance stored in the `mmeth` member variable. As a consequence setting the caller after construction had no effect and could trigger the "You forgot to use setCaller()!" assertion ([LocalOperationCaller::checkCaller()](https://github.com/orocos-toolchain/rtt/blob/toolchain-2.9/rtt/internal/LocalOperationCaller.hpp#L284)) in 2.9 or break callbacks in earlier versions of RTT.

The new unit tests are not directly related, but verify that operations called via a `RemoteOperationCaller` are not called too often, which was observed in an application using toolchain-2.8, and that operations and operation callers with reference arguments preserve the address of their arguments. The latter can be dangerous in case of operations sent to another thread and I am not sure if people are aware of this potential problem or if this is documented somewhere. Maybe it would be better to adapt the signature when creating `RTT::Operation<Signature>` instances from function pointers implicitly, e.g. in [RTT::TaskContext::addOperation(const std::string name, Signature *func, ExecutionThread et)](https://orocos-toolchain.github.io/rtt/master/api/html/classRTT_1_1TaskContext.html#ab20c2fe74dd0691f7c07e72948738c8f) to not run into thread-safety problems if `Signature` contains reference arguments?

### DIFF
--- a/tests/remote_method_test.cpp
+++ b/tests/remote_method_test.cpp
@@ -30,6 +30,14 @@
 #include "unit.hpp"
 #include "operations_fixture.hpp"
 
+#define wait_for_equal( a, b, times ) do { \
+    bool wait_for_helper; \
+    int wait = 0; \
+    while( (wait_for_helper = ((a) != (b))) && wait++ != times ) \
+        usleep(100000); \
+    if (wait_for_helper) BOOST_CHECK_EQUAL( a, b ); \
+} while(0)
+
 /**
  * This test suite tests the RTT::internal::RemoteOperationCaller class
  * and its dependencies, being OperationCallerC and SendHandleC.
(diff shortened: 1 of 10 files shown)


tests/remote_method_test.cpp
```diff
@@ -312,7 +320,76 @@ BOOST_AUTO_TEST_CASE(testRemoteOperationCallerFactory)
...
+    BOOST_CHECK( returnAddressOf.ready() );
+
+    int var = 0;
+
+    // call
+    BOOST_CHECK_EQUAL(&var, returnAddressOf(var));
+    BOOST_CHECK_EQUAL(&var, returnAddressOfConst(var));
+
+    // send
+    // Collecting from a RemoteOperationCaller returning void * segfaults!
```
meyerj: This might be another bug and I have not found the root cause yet. There is a comment about RemoteOperationCaller not supporting reference return types in [rtt/internal/DataSourceStorage.hpp:133](https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/DataSourceStorage.hpp#L133), so maybe this also affects pointer return types.

Issue #149, opened 2016-04-26
importing peers in functions


ahoarau:
The following code causes an error  :
```ruby
global void import_stuff()
{
   import("rtt_ros") # creates "this.ros"
   ros.import("test")
}
```

`deployer -s import_test.ops` : 

`import_test.ops :Parse error at line 5: Service or Task "Deployer" has no Peer or Service ros (or Deployer was not found at all).`

But this does **not** :
```ruby
global void import_stuff()
{
   import("rtt_ros")
   scripting.eval("ros.import(\"test\")")
}
```
Whenever a function ran into runScript, it does not accept newly created services/peers.

ahoarau:
@meyerj gentle ping on this issue.

meyerj:
There is not much you can do about that without refactoring the whole scripting engine. Orocos does not evaluate every line of a function or program on the fly. In a first step it parses the function and already creates all the necessary data sources, operation callers etc., in order to be able to execute it with real-time constraints (under some additional assumptions and depending on the data types involved). At the time the parser sees the line `ros.import("test")` it has to know that `ros` resolves to a service registered globally and that this service provides an operation named `import` with a certain signature, but the function has not been executed yet and hence the `ros` service does not exist.

It might be possible to automatically delay the evaluation of a line until the function is actually executed in case of an error during the initial parsing step, so basically to interpret the first definition of `import_stuff()` like in the second in your example. But that does not appear to be a sensible approach to me if real-time execution really matters and might hide potential problems it the function is really called repeatedly in a real-time context.

Your use case looks like `import_stuff()` only needs to be called once during deployment, but ROS is optional in your application and you do not want to import package `rtt_ros` unconditionally, outside the function? In that case you could move the body of the function into a separate file `import_stuff.ops` and "call" it with `scripting.runScript("import_stuff.ops")` instead?

ahoarau:
I'm trying to get to a sort-of-dynamic deployment, where you load a big orocos lib that allows users to create components in one line (instead of the whole loadComponent, setActivity, configure, start, connect to the provided component etc). Something like : 

```c#
import("rtt_rospack")
ros.import("components_lib.ops") // <-- Defines loadComp1(), loadComp2() etc
loadComp1()
``` 

And ```loadComp1()``` is defined like : 

```c#
global void loadComp1()
{
   loadComponent("hello","OCL::HelloWorld")
   hello.do_stuff()
}
``` 

[...]

PR #249, opened 2017-12-07
Add support for peers names as string variables

ahoarau:
This PR allows to write scripts as :
```c
var string comp_name = "hello"
loadComponent(comp_name,"OCL::HelloWorld")
comp_name.configure()
comp_name.start()
```

### DIFF
--- a/rtt/scripting/PeerParser.cpp
+++ b/rtt/scripting/PeerParser.cpp
@@ -169,8 +169,24 @@ namespace RTT
 
     void PeerParser::locatepeer( iter_t begin, iter_t end )
     {
-        std::string name( begin, end );
-        name.erase( name.length() -1  ); // compensate for extra "."
+        std::string name_tmp( begin, end );
+        name_tmp.erase( name_tmp.length() -1  ); // compensate for extra "."
+        std::string name; // the peer name
+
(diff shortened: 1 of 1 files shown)


meyerj:
See also my comment in https://github.com/orocos-toolchain/rtt/issues/149#issuecomment-355122058 with an alternative proposal to solve the problem.

I do not like the implicit conversion from string variables to a peer very much and would suggest the opposite approach, namely to return a `TaskContext *` from `loadComponent()`. As a result, your example would have to be written like
```c
var string comp_name = "hello"
var TaskContext comp = loadComponent(comp_name,"OCL::HelloWorld")
comp.configure()
comp.start()
```
with a clear separation between the actual component (`comp`) and its name (`comp_name`). Additionally, each TaskContext could expose `getPeer(string name)` as an operation to convert from names to actual TaskContext instances explicitly.

Implicit conversion from `TaskContext` to `string` or vice-versa as originally proposed in https://github.com/orocos-toolchain/rtt/issues/149#issuecomment-355122058 might not even be required if the PeerParser handles variables of type `TaskContext` in the same way as peers, so that names of peers and variables pointing to TaskContexts can be used interchangeably in scripting.

PR #254, opened 2018-01-05
update gcc version check

vbargsten:
...to allow gcc version 7 on Ubuntu 17.10. The original check allowed versions up to 6; however, the error message was quite confusing since it was out of sync (requiring 3.x or 4.x) as well as the status message. CXX_VERSION is detected as:
```
$ g++ -dumpversion
7
```
Thus the match won't work anymore as it requires a minor version number. Match was changed to also force beginning of string; otherwise e.g. 3.5 will be matched anywhere (for example in 8.3.5). Check reduced to warning level.

Btw, if actually someone would still use gcc 3, he could run into trouble. I tested RTT_GCC_HASVISIBILITY FALSE (with the 7 version) and we get some duplicate definitions (RTT_EXT_IMPL empty then):
```
...rtt/rtt/rt_string.cpp:45:20: error: duplicate explicit instantiation of 'class std::__cxx11::basic_string<char, std::char_traits<char>, RTT::os::rt_allocator<char> >' [-fpermissive]
...rtt/rtt/rt_string.cpp:46:20: error: duplicate explicit instantiation of 'class std::__cxx11::basic_ostringstream<char, std::char_traits<char>, RTT::os::rt_allocator<char> >' [-fpermissive]
```
from `grep -rin basic_string rtt/`
```
rtt/rt_string.cpp:45:    template class basic_string<char, char_traits<char>, RTT::os::rt_allocator<char> >;
rtt/rt_string.hpp:50:    typedef std::basic_string<char, std::char_traits<char>, RTT::os::rt_allocator<char> > rt_string;
rtt/rt_string.hpp:74:    RTT_EXT_IMPL template class basic_string<char, char_traits<char>, RTT::os::rt_allocator<char> >;
```

### DIFF
--- a/config/check_depend.cmake
+++ b/config/check_depend.cmake
@@ -317,6 +317,10 @@ ENDIF(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "
 # If we're using gcc, make sure the version is OK.
 #
 IF (CMAKE_COMPILER_IS_GNUCXX)
+  
+  # default for version >= 4
+  SET(RTT_GCC_HASVISIBILITY TRUE)
+  
   # this is a workaround distcc:
   IF ( CMAKE_CXX_COMPILER_ARG1 )
     STRING(REPLACE " " "" CMAKE_CXX_COMPILER_ARG1 ${CMAKE_CXX_COMPILER_ARG1} )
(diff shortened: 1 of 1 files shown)


meyerj:
Thanks for your patch and sorry for the inconvenience, but this is already at least the 3rd issue or pull request on this issue. I would propose to merge https://github.com/orocos-toolchain/rtt/pull/200 instead if there are no objections. The open TODO items could be split into a separate issue.

vbargsten:
Sounds good, less code to maintain. I think it is reasonable to merge [#200](https://github.com/orocos-toolchain/rtt/pull/200) and drop support for gcc 3.

PR #138, opened 2016-02-09
gnulinux: use CLOCK_MONOTONIC for periodic tasks (fix #105)

meyerj:
This PR applies the patch proposed by @rbrtslmn in #105.

It switches to use a monotonic clock source for periodic waiting in the gnulinux target, but it is also somehow incomplete because it does not fix the same issue for other broken targets (like MacOS), for waiting on condition variables and for RTT Timers.

PR #258, opened 2018-03-01
Use a monotonic clock source whenever possible (gnulinux target)

### LINKED ISSUE #233: Consistently use a steady clock source for periodic tasks and timers
From https://github.com/orocos-toolchain/rtt/pull/129#issuecomment-338789418:
> I am thinking about reverting this patch for toolchain-2.9 in favor of https://github.com/orocos-toolchain/rtt/pull/138. #138 only provides a solution for the `Thread` class, but not for `Timer`.

> Also it should be noted that both patches, #129 and #138, are not effective anymore for periodic components in connection with #91. So actually we would need a clean solution for all these cases of interruptible, perodic waiting by adding an OS abstraction of condition variables which support waiting based on a steady clock source. For gnulinux/pthread that would be using [pthread_condattr_setclock](http://pubs.opengroup.org/onlinepubs/9699919799/) with `CLOCK_MONOTONIC`. And get rid of the semaphore in `os::Timer` by using a condition variable instead.

I would actually consider the fact that RTT cannot yet handle this correctly as a serious bug. I think the xenomai target does already use a monotonic clock source for time services, condition variables and semaphores, but not gnulinux and macosx.

Thoughts?

meyerj:
Fixes #233.

The absolute time returned by `rtos_get_time_ns()` or `RTT::os::TimeService::getNSecs()` was generated from a wall clock (time of day), which is not monotonic and can jump in certain cases, e.g. due to an NTP update. This time source is used in several internal and public API functions to implement timers or to wait on a condition variable, mutex or semaphore with a timeout. In all these cases non-steady system time can have undesirable side effects, which are not expected in a Real-Time Toolkit.

There have been several attempts to fix or at least mitigate that problem:
- http://bugs.orocos.org/show_bug.cgi?id=735
- #129 
- #138

This patch introduces some potentially breaking semantic changes, with only minimal changes in the public API:

- Use `clock_gettime(CLOCK_MONOTONIC, ...)` to implement `rtos_get_time_ns()` for the gnulinux target.

  This change has a major impact on all usages of that method, `rtos_get_ticks()` or the higher-level methods `RTT::os::TimeService::getNSecs()` and `RTT::os::TimeService::getTicks()`, whose return values cannot be interpreted as wall time anymore.

  It should be noted that this was never promised by the RTT API and the `xenomai` target already uses another clock source which is not related to wall time.

- Consequently remove the `rtos_get_time_monotonic_ns()` function introduced in #138. `rtos_get_time_ns()` is monotonic now.

- Add new OS abstraction functions for waiting on a mutex lock with a relative timeout and use them      to implement `RTT::os::Mutex::timedlock()`.

  POSIX does not provide a `pthread_mutexattr_setclock()` function similar to `pthread_condattr_setclock()`. For that reason `rtos_mutex_lock_until()` and `rtos_mutex_trylock_for()` have to fall back to `CLOCK_REALTIME` and may still be affected by time adjustments while waiting.

  Another possible workaround would be to pair the mutex with a condition variable which is notified whenever a thread releases the mutex, so that threads waiting to acquire the mutex can wait on the condition variable instead. This is also the approach taken to implement `boost::timed_mutex` in [boost/thread/pthread/mutex.hpp](https://github.com/boostorg/thread/blob/96484ac57fc0406391782a14387f56361e63f2f0/include/boost/thread/pthread/mutex.hpp#L307-L315).

- Drop `RTT::os::Semaphore::waitUntil()` and the underlying OS abstraction functions.

  The only usage within the RTT core was in RTT::os::Timer, which can be rewritten using a condition variable. POSIX semaphores do not support waiting with an absolute timeout from the monotonic clock.

  If still required by some user code, semaphores could be augmented by an additional condition variable  in the same way as suggested for mutexes.

- Revert #129, which was a workaround and should not be required anymore with this patch.

The `macosx` target is affected, too. I tried to replace `rtos_get_time_ns()` by [mach_absolute_time()](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time), but because Orocos uses the Boost implementations for mutexes and condition variables, which do not support another clock source than CLOCK_REALTIME, there is no easy solution. Suggestions are welcome.

I only tested this patch by running unit tests with targets `gnulinux` and `macosx` so far and verified that compilation succeeds for the `xenomai` target.

The last patch https://github.com/orocos-toolchain/rtt/commit/02ff20681aebbe8b24e110804ea9e11483d925d0 was cherry-picked from toolchain-2.9 and is required to compile successfully for newer Mac OS X versions.


jwillemsen:
For the CORBA Integration TAO can also be configured to use monotonic time, to ace.config.h the following lines have to be added:
```
#define TAO_USE_HR_TIME_POLICY_STRATEGY
#define ACE_HAS_CLOCK_GETTIME_MONOTONIC

```

meyerj:
> For the CORBA Integration TAO can also be configured to use monotonic time, to ace/config.h the following lines have to be added:

Thanks for the info. This appears to be a built-time configuration option in TAO. Or would you expect that RTT defines these preprocessor macros somewhere before including TAO headers?

meyerj:
@doudou, @pdima, @rbrtslmn and others: Any objections against changing `rtos_get_time_ns()` and `RTT::os::TimeService::getNSecs()` to return a timestamp from a monotonic clock and dropping `RTT::os::Semaphore::waitUntil()` on `master` and `toolchain-2.9`?

jwillemsen:
`ACE_HAS_CLOCK_GETTIME_MONOTONIC` is a mandatory compile time option to enable the monotonic time, `TAO_USE_HR_TIME_POLICY_STRATEGY` is an optional one, it can also done through the commandline but it is safer/easier to just let the ORB use a monotonic time through the define

meyerj:
> ACE_HAS_CLOCK_GETTIME_MONOTONIC is a mandatory compile time option to enable the monotonic time, TAO_USE_HR_TIME_POLICY_STRATEGY is an optional one, it can also done through the commandline but it is safer/easier to just let the ORB use a monotonic time through the define

What I meant is that this option already has to be defined when ACE/TAO is built, not RTT. Or would you suggest to patch [rtt/transports/corba/ApplicationServer.cpp](https://github.com/Intermodalics/rtt/blob/e95e9da3e95de3e287bf773d15b83e2e2f02977e/rtt/transports/corba/ApplicationServer.cpp) in any way?

jwillemsen:
Yes, these options have to be set in the ace/config.h file which has to be done before ACE/TAO are compiled

doudou:
@meyerj no objection on my side

meyerj:
Thanks for your feedback!

snrkiwi:
Is there any possible interaction between things that wait on the monotonic clock, and those that wait on the realtime clock?

Did I miss any documentation updates that _rtos_get_time_ns()_ is now monotonic and based on an arbitrary epoch, while _rtos_get_realtime_ns()_ is non-monotonic, subject to NTP adjustments but uses a wall-clock useful epoch?

This is definitely a backwards incompatible change for any application that uses anything related to RTT time services, and compares or relates it to any other wall clock source. We should make sure that that is understood.

meyerj:
> Is there any possible interaction between things that wait on the monotonic clock, and those that wait on the realtime clock?

Not within RTT unless I overlooked something. The only usage within RTT is for `rtos_mutex_[rec_]trylock_for()`, because pthread mutexes do not support waiting with a relative timeout or with an absolute time using `CLOCK_MONOTONIC` (reference: https://stackoverflow.com/questions/14248033/clock-monotonic-and-pthread-mutex-timedlock-pthread-cond-timedwait). Sticking to a monotonic clock source whenever possible is in general a good thing for a real-time framework.

> Did I miss any documentation updates that rtos_get_time_ns() is now monotonic and based on an arbitrary epoch, while rtos_get_realtime_ns() is non-monotonic, subject to NTP adjustments but uses a wall-clock useful epoch?

[rtos_get_time_ns()](https://github.com/orocos-toolchain/rtt/pull/258/files#diff-39403cc75fc553eb965d446766e58f07R100) and [rtos_get_realtime_ns()](https://github.com/orocos-toolchain/rtt/pull/258/files#diff-39403cc75fc553eb965d446766e58f07R108) are considered internal and were not documented before. We could add a clarification to the [TimeService](https://github.com/Intermodalics/rtt/blob/e95e9da3e95de3e287bf773d15b83e2e2f02977e/rtt/os/TimeService.hpp) API description and state explicitly that the clock source is not necessarily a wall clock, but on the other hand this was never promised. For targets other than `gnulinux` and `macosx`, that's also not the case.

Currently there is no OS-independent way in RTT to retrieve a wall clock timestamp in a real-time context, e.g. for message headers. Actually there never was. For the `gnulinux` target the new internal `rtos_get_realtime_ns()` should be real-time safe, Xenomai's POSIX skin provides a real-time safe `clock_gettime()` implementation. If we would add such a function, it should probably not return a `long long` nanosecond counter, but a `struct timespec` data structure to clearly distinguish between monotonic RTT time and real time and to avoid that users compare one with the other.

> This is definitely a backwards incompatible change for any application that uses anything related to RTT time services, and compares or relates it to any other wall clock source. We should make sure that that is understood.

I agree. It's definitely worth a section in the pending 2.9 release notes [here](https://github.com/orocos-toolchain/orocos_toolchain/blob/toolchain-2.9/orocos_toolchain/CHANGELOG.rst). Will add that throughout the day.

PR #261, opened 2018-03-03
Fix collectIfDone() for more than three arguments

meyerj:
`collectIfDone(...)` was calling `cimpl->collect(...)` for the specializations with more than three arguments.
Also reordered template specializations in `Collect.hpp`.

This is a nasty one.


snrkiwi:
WFM

PR #271, opened 2018-09-12
Update README.md, made a link to the official Installation Guide for 2.9

qkx515:
The old README is deprecated. The newest one should be the first thing users see to help them install rtt easily.


meyerj:
Thanks for the update. Indeed the old `README` was outdated. The referenced installation manual at https://orocos-toolchain.github.io/rtt/toolchain-2.9/xml/orocos-installation.html is also partially out of date, but still better to reference this one than to replicate the manual in a `README`.

Also note that we once started to create a new installation manual and script (for non-ROS users) at the Orocos Toolchain level in https://github.com/orocos-toolchain/orocos_toolchain/pull/13, but there are still some "to be copied from..." sections. The ultimate goal is to update and convert all the documents from https://github.com/orocos-toolchain/rtt/tree/master/doc to MarkDown or reStructuredText and host it using [GitHub pages](https://pages.github.com/) (like https://orocos-toolchain.github.io) or [Read The Docs](https://readthedocs.org/). This could also replace the Orocos wiki on http://www.orocos.org/, which still has some useful, but mainly outdated content.

Unfortunately there is no time at the moment to follow up on this, but any contribution towards that goal is highly appreciated.

qkx515:
Could you please look at https://github.com/orocos-toolchain/orocos_toolchain/issues/30 and give me a little hint. 

Now I succeed in installing rtt following its installation guide, then log4cpp following rtt's instruction, then ocl following rtt's instruction. But when I tried to create my first package, HelloWorld. It seems I still need to install orogen. And that's where I got stuck. Installing these packages are really painful, I would really love to install orocos-toolchain as a whole.

I will write down a documentation about how to install them and help others, if I succeed! Thank you in advance! :-)

@meyerj

PR #273, opened 2018-11-16
core: add sanity checks in case someone calls exception() in some Hooks

doudou:


meyerj:
Looks good to me. Thanks!

PR #277, opened 2019-04-09
Patch for 2.9.1 ABI compatibility with 2.9.0

francisco-miguel-almeida:
Patch for for the RTT `toolchain-2.9.1` branch that makes it ABI compatible with RTT 2.9.0.


meyerj:
It should be noted that this patch will never end up in master. It was only made to retain ABI-compatibility for the 2.9.1 release to include #215.

PR #278, opened 2019-04-11
tests: fixed number of workers in buffers_test/testListLockFree

meyerj:
This fixes a regression from 75745d6aec03a4fe6cc1a11bfa4362001296fc19: The `ListLockFree` instance was instantiated with a maximum number of threads of 4, but the test case ran 5 workers and one additional grower threads. The number of workers has now been reduced to 3 again, like before 75745d6aec03a4fe6cc1a11bfa4362001296fc19.

The fixture class `BuffersAtomicMWMRQueueTest` was actually not used for the last test cases. Only the `listlockfree` member, which has been moved into the `testListLockFree()` test case.

This bug caused occasional build errors on Travis CI, e.g. https://travis-ci.org/orocos-toolchain/rtt/builds/518819992.

### DIFF
--- a/tests/buffers_test.cpp
+++ b/tests/buffers_test.cpp
@@ -89,17 +89,14 @@ public:
     MWMRQueueType* aqueue;
     ThreadInterface* athread;
     ThreadInterface* bthread;
-    ListLockFree<Dummy>* listlockfree;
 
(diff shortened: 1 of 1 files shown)


francisco-miguel-almeida:
LGTM

PR #279, opened 2019-04-15
Testing xenial build on Travis (DO NOT MERGE)

francisco-miguel-almeida:
PR for testing xenial build on Travis. Once we have the .yml correct, we can cherry pick .travis.yml to `master` and both release branches.


meyerj:
Please rebase this patch on `master`, and not on the `toolchain-2.9.1` release candidate branch that we will never merge back into `master` anyway.

See also the related PR https://github.com/orocos-toolchain/rtt/pull/195, which was an attempt to switch to containerized builds and also compile with different compilers. Unfortunately this build infrastructure has been (deprecated again in the meantime](https://docs.travis-ci.com/user/reference/trusty/#container-based-infrastructure).

I think the best option would be to follow-up in #195 and do the switch to `xenial` there.

Issue #161, opened 2016-07-22
Inconsistent RTT_COMPONENT_PATH overlay semantics


ahoarau:
When I import a component from an overlay workspace it always takes the **last** on the ROS_PACKAGE_PATH list. The order of priority is reversed !
- workspace 1 contains component A
- workspace 2 extends workspace 1, and also contains a different copy of component A

if I load component A (`import("A")`) it will output : 

``` bash
# blabla loading from component A from workspace2 --> success !
# but then, loading component A from workspace 1 (why ??) --> success
 [ Warning][ComponentLoader::import(path_list)] Component type name A already used: overriding.
```

[This](https://github.com/orocos-toolchain/rtt/commit/fb0ba650253ce6372b4a4d9ebd680bf253886aa3) seems to be the culprit.

meyerj:
This is a known problem and RTT can simply not deal with overlay workspaces that provide the same component types or plugins with the same name at the moment.

I somehow remember the discussion back in 2014 about whether to reverse paths in the `RTT_COMPONENT_PATH` or not, but I cannot find a reference now. The real problem is that the ComponentLoader, the PluginLoader and the TypeInfoRepository, the classes who load the plugins or hold information about known components, services and typekits, have inconsistent behaviors. Without reversing the entries in the `RTT_COMPONENT_PATH` given as `"/path/to/overlay:/path/to/underlay"`...

- component libraries loaded later override existing component type names (see @ahoarau's example) --> *prefer underlay*
- service plugins providing the same service can be loaded at the same time, but when loading a service by name the first factory function found is called --> *prefer overlay*
- typekit plugins that provide the same type can add alias names to types loaded earlier, but the factories installed in [TypeInfoGenerator::installTypeInfoObject(TypeInfo *)](https://github.com/orocos-toolchain/rtt/blob/fb0ba650253ce6372b4a4d9ebd680bf253886aa3/rtt/types/TypeInfoGenerator.hpp#L45) (e.g. in [TemplateTypeInfo<T>](https://github.com/orocos-toolchain/rtt/blob/fb0ba650253ce6372b4a4d9ebd680bf253886aa3/rtt/types/TemplateTypeInfo.hpp#L109)). This behavior is closer to *prefer underlay* unless the typekits are fully compatible or complementary.
- transport plugins that provide a transport for the same type and with the same id cannot override a transport registered earlier --> *prefer overlay*

[...]

PR #288, opened 2019-04-30
Change component loader behaviour to consider a component found on first valid path

francisco-miguel-almeida:
Fixes issue #161.

### DIFF
--- a/rtt/deployment/ComponentLoader.cpp
+++ b/rtt/deployment/ComponentLoader.cpp
@@ -385,39 +385,6 @@ bool ComponentLoader::import( std::string const& path_list )
             // If the path is not complete (not absolute), look it up in the search directories:
             log(Debug) << "No such directory: " << p<< endlog();
         }
-#if 0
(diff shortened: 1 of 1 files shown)


meyerj:
For reference: This PR implements what has been proposed in https://github.com/orocos-toolchain/rtt/issues/161#issuecomment-242874748.

> It would be an option to simply load a package from the first path found in [ComponentLoader::importInstalledPackage()](https://github.com/orocos-toolchain/rtt/blob/master/rtt/deployment/ComponentLoader.cpp#L429). This should fix the overlay semantics of the `import()` operation and would not break the component or typekit overwrite behavior of a subsequent import call of another package.

So if a package exists **with the same name** in multiple directories in the `RTT_COMPONENT_PATH` (e.g. catkin overlay and underlay), only the first one is imported (prefer overlay). If packages **with different names** provide component types, service plugins or Orocos types with conflicting names (e.g. name clashes or renamed package in one workspace), the inconsistent overlay semantics described in https://github.com/orocos-toolchain/rtt/issues/161#issuecomment-234884716 still apply.

PR #293, opened 2019-05-16
Fix multi-signal emission in case of fused functor data source callbacks

meyerj:
This is a bunch of patches in `toolchain-2.9` that have not yet been on any pull request so far:

From https://github.com/orocos-toolchain/rtt/commit/619fce27b36bfb5abaaeb68ba9ba8fb1683747aa (by @psoetens):
> signals: fix multi-signal emission in case of fused functor data source callbacks
> 
> When an operation was called twice in a row, and a signal callback was installed
using datasource semantics (with produceSignal() ), both callbacks would most
likely receive twice the same data, being equal to the last operation call's values,
since before the callbacks are dispatched, the data sources are already filled in.
The last write wins and the first call-back sees the data of the last.
> 
> This patch fixes that by defining a data_store_type in create_sequence with a
store() and a load() function. The fused signal callback can now first store
the args of the operation, then dispatch and retrieve them later on in the
callback function.

A later patch removed the new `DataStore<T>` class defined in [CreateSequence.hpp](https://github.com/orocos-toolchain/rtt/blob/ca0c42bab65f2696a304ea6935fda13a902e1c6e/rtt/internal/CreateSequence.hpp) again. Instead we use `AStore<T>` already defined in [BindStorage.hpp](https://github.com/orocos-toolchain/rtt/blob/ca0c42bab65f2696a304ea6935fda13a902e1c6e/rtt/internal/BindStorage.hpp) involved in operation calls.

The last patch, https://github.com/orocos-toolchain/rtt/commit/ca0c42bab65f2696a304ea6935fda13a902e1c6e, removes a `using namespace std` directive in [FusedFunctorDataSource.hpp](https://github.com/orocos-toolchain/rtt/blob/ca0c42bab65f2696a304ea6935fda13a902e1c6e%5E/rtt/internal/FusedFunctorDataSource.hpp#L58:L59), probably a leftover from debugging https://github.com/orocos-toolchain/rtt/commit/619fce27b36bfb5abaaeb68ba9ba8fb1683747aa. This commit was cherry-picked directly to `toolchain-2.9` as 23f826a95b986641c70226f9ee50874a4f0d201e.

### DIFF
--- a/rtt/internal/BindStorage.hpp
+++ b/rtt/internal/BindStorage.hpp
@@ -289,7 +289,7 @@ namespace RTT
             typename Signal<ToBind>::shared_ptr msig;
 #endif
 
-            BindStorageImpl() :  vStore(boost::ref(retv)) {}
+            BindStorageImpl() :  vStore(retv) {}
             BindStorageImpl(const BindStorageImpl& orig) : mmeth(orig.mmeth), vStore(retv)
 #ifdef ORO_SIGNALLING_OPERATIONS
                                                          , msig(orig.msig)
@@ -337,10 +337,10 @@ namespace RTT
             void store(arg1_type t1) { a1(t1); }
             void exec() {
 #ifdef ORO_SIGNALLING_OPERATIONS
-                if (msig) (*msig)(a1.get());
+                if (msig) (*msig)(a1);
 #endif
                 if (mmeth)
-                    retv.exec( boost::bind(mmeth, boost::ref(a1.get()) ) );
+                    retv.exec( boost::bind(mmeth, a1 ) );
                 else
                     retv.executed = true;
             }
@@ -377,10 +377,10 @@ namespace RTT
             void store(arg1_type t1, arg2_type t2) { a1(t1); a2(t2); }
             void exec() {
 #ifdef ORO_SIGNALLING_OPERATIONS
-                if (msig) (*msig)(a1.get(), a2.get());
+                if (msig) (*msig)(a1, a2);
 #endif
                 if (mmeth)
-                    retv.exec( boost::bind(mmeth, boost::ref(a1.get()), boost::ref(a2.get()) ) );
+                    retv.exec( boost::bind(mmeth, a1, a2 ) );
                 else
                     retv.executed = true;
             }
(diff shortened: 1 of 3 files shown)


rtt/internal/BindStorage.hpp
```diff
@@ -64,27 +64,29 @@ namespace RTT
...
+            typedef T arg_type;
             T arg;
             AStore() : arg() {}
             AStore(T t) : arg(t) {}
             AStore(AStore const& o) : arg(o.arg) {}
 
             T& get() { return arg; }
             void operator()(T a) { arg = a; }
-            operator T() { return arg;}
+            operator T&() { return arg; }
```
meyerj: @psoetens Was there a good reason why the operator did not return a reference in the original code? I think not: The object is already copied at least once when creating the `AStore<T>` instance or via `operator()(T a)`. So no reason to copy it again to read it. For references we use the specialization below instead.

Even without the reference return type the object was not copied for the actual operation call because before this patch `.get()` was used to access it in `BindStorageImpl<ToBind>` below.

rtt/internal/BindStorage.hpp
```diff
@@ -335,10 +337,10 @@ namespace RTT
             void store(arg1_type t1) { a1(t1); }
             void exec() {
 #ifdef ORO_SIGNALLING_OPERATIONS
-                if (msig) (*msig)(a1.get());
+                if (msig) (*msig)(a1);
 #endif
                 if (mmeth)
-                    retv.exec( boost::bind(mmeth, boost::ref(a1.get()) ) );
+                    retv.exec( boost::bind(mmeth, a1 ) );
```
meyerj: To be checked: Does `boost::bind` actually copies the `AStore<T>` instance and indirectly the argument to the functor object or does it invoke `operator T&()` here and only stores a reference?

meyerj: Seems like [boost::bind]() is always copying the arguments itself, and only invokes the conversion operator at the time the functor is actually called: https://stackoverflow.com/questions/11255144/why-does-boostbind-store-arguments-of-the-type-passed-in-rather-than-of-the-ty

In this case the patch in a5857b2aab16144ed5ddb314209529de0d06cc9c (this PR) and a2ccd7961975a80f4befd699a7edc8b029085cea (in `toolchain-2.9`) is actually wrong. The `.get()` can be removed, but the `boost::ref()` call should have been kept.

rtt/internal/BindStorage.hpp
psoetens: I remember that I tested this code with objects which did a cout in a copy constructor to count the number of copies made, and then introduced references to reduce the number of copies. That said, I don't remember if there was a formal place in the architecture where a copy 'must' be made, but the AStore seems to be the best place to do it, if required. AStore was meant to mimic the classical C++ function argument behavior.

PR #295, opened 2019-05-22
Fix compiler warnings for discarded return values in tlsf

francisco-miguel-almeida:
As described in title

### DIFF
--- a/rtt/os/tlsf/tlsf.c
+++ b/rtt/os/tlsf/tlsf.c
@@ -377,9 +377,10 @@ static __inline__ bhdr_t *FIND_SUITABLE_BLOCK(tlsf_t * _tlsf, int *_fl, int *_sl
 }
 
 static __inline__ void corrupt(const char *msg) {
-    static const char *k =  "* Heap corruption detected: *\n";
-    write( STDERR_FILENO, k, strlen(k) );
-    write( STDERR_FILENO, msg, strlen(msg) );
+#ifdef USE_PRINTF
+    fputs("* Heap corruption detected: *\n", stderr);
+    fputs(msg, stderr);
+#endif
     abort();
 }
 


rtt/os/tlsf/tlsf.c
```diff
@@ -376,13 +376,10 @@ static __inline__ bhdr_t *FIND_SUITABLE_BLOCK(tlsf_t * _tlsf, int *_fl, int *_sl
...
 
-#define IGNORE_RETURN_VALUE(var, expr) var = (expr); (void) var
-
 static __inline__ void corrupt(const char *msg) {
     static const char *k =  "* Heap corruption detected: *\n";
-    int _ignored;
-    IGNORE_RETURN_VALUE(_ignored, write( STDERR_FILENO, k, strlen(k) ));
-    IGNORE_RETURN_VALUE(_ignored, write( STDERR_FILENO, msg, strlen(msg) ));
+    fputs(k, STDERR_FILENO);
+    fputs(msg, STDERR_FILENO);
```
meyerj: The second argument is not a file descriptor (`STDERR_FILENO`), but a `FILE *` (`stderr`).

francisco-miguel-almeida: The test also failed (locally) when I corrected that to use the appropriate `stderr` file pointer. I just fixed it by using a `void` cast with negation instead.

meyerj: Works for me locally. I will push a new commit and it Travis is happy, I suggest to merge the `fputs()` version.

It is normal that the `tlsf_test` outputs heap corruption errors. That is the expected behavior:
```
Running 6 test cases...
* Heap corruption detected: *
free_ex(): Freeing unused block
```

Issue #300, opened 2019-06-14
Massive deadlocking when disconnecting connections in parallel


doudou:
The refactored channel implementation that just landed in master has a very liberal use of locking, which leads to deadlocks when doing parallel disconnections. I'm chasing those, and will propose a PR for review, I'm opening this issue to track the problem.

Generally speaking, one major issue I see (apart from the deadlocking) is that remote disconnection calls are done under lock. My current strategy is to split the "remove channel from channel list(s)" from the "destroy the channel".

meyerj:
Thanks for reporting. Could you already share an example or unit test to reproduce the problem?

There is a set of new test cases in [ports_test.cpp](https://github.com/orocos-toolchain/rtt/blob/809e15cfa442a3ad31ae1158fc1a8a38cb9f9737/tests/ports_test.cpp#L1245), introduced in https://github.com/orocos-toolchain/rtt/commit/f1404ff714d3bee75b1c560fa3132cc9108b139c, that should have covered issues with parallel port connection, disconnection, reads and writes. But indeed it only spawns a single thread to add and remove an input port from the connection and another for an output port, but never two input or output ports concurrently.

What you suggest sounds a bit like (partially) reverting https://github.com/orocos-toolchain/rtt/pull/283, a patch that was only added recently. Without, the test cases in `ports_test.cpp` mentioned above did not check whether the port connections were actually successful, which was often not the case.

doudou:
See #302

PR #302, opened 2019-06-17
[WIP] fix deadlock in concurrent connection disconnection

doudou:
This PR aims at fixing the issue identified in #300. The deadlock happens essentially when disconnecting the same ports from two directions. The easiest way is to disconnect the same connection from both directions at the same time, which leads reliably to deadlocking (associated test in https://github.com/orocos-toolchain/rtt/commit/bb6fbc14306362e1e92c1160c58058bc2492ac34).

So far, the PR fixed the few easy places. However, there are a few very large codepaths still executed under lock, as removing the lock was breaking the tests without an obvious reason from my side. I figured that you guys might have more insight.

- [ ] the whole connection creation codepath is done under lock (for instance https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/ConnFactory.cpp#L216, https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/ConnFactory.hpp#L460). The original design was making sure that the channel is created first and then added to the ports, which would remove the need for such a wide-ranging lock.
- [ ] the endpoint disconnection is done under lock for the port. I am also not sure why, as the connection manager is meant to be thread-safe already. But tests start failing without that lock, so it probably is there for a good reason: https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/ConnOutputEndPoint.hpp#L133 and https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/ConnInputEndPoint.hpp#L76.

In general, my gut feeling is that the general design should be to connect/disconnect a channel from a port under the port lock, but wait until the lock is released to actually dismantle the channel. This would ensure that the large latency part (hitting the transports) is done without affecting the component.

### DIFF
--- a/rtt/base/ChannelElement.hpp
+++ b/rtt/base/ChannelElement.hpp
@@ -98,7 +98,7 @@ namespace RTT { namespace base {
         }
 
         /** Writes a new sample on this connection. \a sample is the sample to
-         * write. 
+         * write.
          *
          * @returns false if an error occured that requires the channel to be invalidated. In no ways it indicates that the sample has been received by the other side of the channel.
          */
@@ -241,6 +241,12 @@ namespace RTT { namespace base {
         {
             if (last == input) last = 0;
             MultipleInputsChannelElementBase::removeInput(input);
+        }
+
+        virtual void removedInputs(Inputs const& inputs)
+        {
+            if (find(inputs.begin(), inputs.end(), last) != inputs.end())
+                last = 0;
         }
 
     private:
--- a/rtt/base/ChannelElementBase.hpp
+++ b/rtt/base/ChannelElementBase.hpp
@@ -248,11 +248,11 @@ namespace RTT { namespace base {
         }
 
         /**
-         * This function may be used to identify, 
+         * This function may be used to identify,
          * if the current element uses a network
          * transport, to send the data to the next
          * Element in the logical chain.
-         * 
+         *
          * @return true if a network transport is used.
          * */
         virtual bool isRemoteElement() const;
@@ -264,7 +264,7 @@ namespace RTT { namespace base {
          * E.g: In the local case output->getLocalURI()
          * In the remote case the URI of the remote
          * channel element.
-         * 
+         *
          * @return URI of the next element.
          * */
         virtual std::string getRemoteURI() const;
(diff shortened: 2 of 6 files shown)


meyerj:
Thanks for your patch!

> * [ ]  the whole connection creation codepath is done under lock (for instance https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/ConnFactory.cpp#L216, https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/ConnFactory.hpp#L460). The original design was making sure that the channel is created first and then added to the ports, which would remove the need for such a wide-ranging lock.

The new problem introduced with #114 is that some buffers might only have to be created once per input port or once per output port. The first implementation of the new dataflow semantics was implemented without that port-wide connection lock, but that caused race conditions when connecting concurrently because the new shared buffer or data object instances were eventually created by two connections, such that one of them would end up in a de-facto disconnected state. This was fixed by https://github.com/orocos-toolchain/rtt/pull/283, which introduced the per-port connection locks.

The `ConnectionManager` actually plays a subordinate now with that patch: It still keeps track of the port's connection, but is not involved at all in the actual dataflow anymore. It is purely for management purposes.

> * [ ]  the endpoint disconnection is done under lock for the port. I am also not sure why, as the connection manager is meant to be thread-safe already. But tests start failing without that lock, so it probably is there for a good reason: https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/ConnOutputEndPoint.hpp#L133 and https://github.com/orocos-toolchain/rtt/blob/master/rtt/internal/ConnInputEndPoint.hpp#L76.

Same issue here: Without that lock it can happen that one thread disconnects the port resulting in the destruction of a shared buffer ([ConnInputEndPoint.hpp:91-95](https://github.com/orocos-toolchain/rtt/blob/809e15cfa442a3ad31ae1158fc1a8a38cb9f9737/rtt/internal/ConnInputEndPoint.hpp#L91:L95)), while another thread creates a new connection. That also could lead to race conditions and a channel pipeline not consistent with the port connections stored in the `ConnectionManager`.

> In general, my gut feeling is that the general design should be to connect/disconnect a channel from a port under the port lock, but wait until the lock is released to actually dismantle the channel. This would ensure that the large latency part (hitting the transports) is done without affecting the component.

I agree. Your patch looks good at first sight. I will do some more tests and check why it [fails to build on Travis](https://travis-ci.org/orocos-toolchain/rtt/builds/546867450).

meyerj (changes requested):

PR #305, opened 2019-10-02
RTT memory leaks

francisco-miguel-almeida:
Possible memory leaks detected with Valgrind, or spotted during code inspection.

### DIFF
--- a/rtt/base/ChannelElement.hpp
+++ b/rtt/base/ChannelElement.hpp
@@ -63,6 +63,8 @@ namespace RTT { namespace base {
         typedef typename boost::call_traits<T>::param_type param_t;
         typedef typename boost::call_traits<T>::reference reference_t;
 
+        virtual ~ChannelElement() { }
+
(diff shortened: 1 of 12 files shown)


rtt/base/ChannelElement.hpp
```diff
@@ -63,6 +63,8 @@ namespace RTT { namespace base {
         typedef typename boost::call_traits<T>::param_type param_t;
         typedef typename boost::call_traits<T>::reference reference_t;
 
+        virtual ~ChannelElement() { }
```
meyerj: The base class already has a virtual destructor and the implicitly defined destructor is therefore also `virtual`. You could add an explicit definition, but it does not change anything related to potential memory leaks and might break the ABI, so I would rather keep those class declarations as-is.

francisco-miguel-almeida: Reverting the (redundant) added explicit destructors, nothing of value is added by those.

PR #309, opened 2019-10-24
FileDescriptorActivity locking and robustness

meyerj:
This PR collects some patches to improve robustness of the `FileDescriptorActivity` and to avoid locks in `trigger()` and `breakLoop()` calls.

- Remove locks from `FileDescriptorActivity::trigger()`, `FileDescriptorActivity::breakLoop()` and `FileDescriptorActivity::triggerUpdateSets()` (contributed by @MagnaboscoL)
- Do not rearm the timeout timer if the previous cycle was due to a `trigger()` call. The previous implementation was waiting for the set timeout period after every interruption, which in theory could lead to the situation that the timeout cycle is never triggered if there is no activity on the file descriptor but the activity is constantly triggered due to operation calls or port callbacks.
- The bookkeeping of the `m_running` flag was buggy and `isRunning()` might have returned false even though the activity was still processing the `step()` function.


rtt/extras/FileDescriptorActivity.cpp
```diff
@@ -319,24 +318,35 @@ void FileDescriptorActivity::loop()
...
-			static const int USECS_PER_SEC = 1000000;
-            timeval timeout = { m_timeout_us / USECS_PER_SEC,
-                                m_timeout_us % USECS_PER_SEC};
+            // only rearm the timer if the previous call was not a pure command pipe event
+            if (m_has_timeout || m_has_ioready || m_has_error ||
+                (timeout.tv_sec == 0 && timeout.tv_usec == 0)) {
+                static const int USECS_PER_SEC = 1000000;
+                timeout.tv_sec = m_timeout_us / USECS_PER_SEC;
+                timeout.tv_usec = m_timeout_us % USECS_PER_SEC;
+            }
```
doudou, line 334: Would be good to add a unit test for this one.

PR #317, opened 2020-03-22
Fix compiler warnings with GCC 9 (with API changes)

meyerj:
This pull request eliminates some compiler warnings with newer compilers (GCC 9 in upcoming Ubuntu 20.04, with Boost 1.72). Some warnings are older and already appeared before.

The patch also includes an API update proposal for more consistent copy assignment semantics of `RTT::Property<T>`, which was inconsistent with copy construction, and a slightly modified API of `RTT::PropertyBag` for adding properties, that likely has no effect in practice.

### DIFF
--- a/rtt/Handle.cpp
+++ b/rtt/Handle.cpp
@@ -53,11 +53,6 @@ namespace RTT {
     {
     }
 
-    Handle::Handle(const Handle& hs)
-        : m_conn( hs.m_conn )
-    {
-    }
-
     Handle::~Handle()
     {
     }
--- a/rtt/Handle.hpp
+++ b/rtt/Handle.hpp
@@ -77,11 +77,6 @@ namespace RTT
         Handle(connection_t conn);
 
         /**
-         * Create a copy-equivalent Handle.
-         */
-        Handle(const Handle& hs);
-
-        /**
          * No-op destructor, does not change signal/slot state.
          */
 		~Handle();
--- a/rtt/Property.hpp
+++ b/rtt/Property.hpp
@@ -174,13 +174,38 @@ namespace RTT
         }
 
         /**
+         * Assignment copies the name, description and value
+         * as \b deep copies.
+         * @post ready() will be true if orig.ready() is true.
+         */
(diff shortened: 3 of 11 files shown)


snrkiwi:
LGTM

meyerj:
I am not so sure yet about the new assignment semantics of `RTT::Property<T>`. I was actually surprised that copy construction and copy assignment had different semantics, but I think it is more consistent like it is now:

- Copying or assigning from another `RTT::Property<T>` of the same type creates a deep copy (value semantics).
- Copying or assigning from a `RTT::base::PropertyBase*` pointer only copies the name and description, but has reference semantics for the actual value. The type of the source property still has to match or the target property becomes invalid (`ready() == false`). The original property and the copy will share the same data source (mirroring).

On the other hand some existing code might depend on the old semantics. RTT unit tests still pass successfully.

snrkiwi:
Probably best we can do. The inconsistencies in application will have to be dealt with ... :-(

meyerj:
`RTT::Attribute` also has the same, more consistent semantics:
- [`Attribute( const Attribute<T>& a)`](https://github.com/orocos-toolchain/rtt/blob/3d8c611863b99fdcbab50f01ebe167b459cf2b54/rtt/Attribute.hpp#L123) creates a deep copy.
- [`Attribute<T>& operator=( const Attribute<T>& a)`](https://github.com/orocos-toolchain/rtt/blob/3d8c611863b99fdcbab50f01ebe167b459cf2b54/rtt/Attribute.hpp#L132) creates a deep copy.
- [`Attribute( base::AttributeBase* ab)`](https://github.com/orocos-toolchain/rtt/blob/3d8c611863b99fdcbab50f01ebe167b459cf2b54/rtt/Attribute.hpp#L150) creates a shallow copy (mirror).
- [`Attribute<T>& operator=(base::AttributeBase* ab)`](https://github.com/orocos-toolchain/rtt/blob/3d8c611863b99fdcbab50f01ebe167b459cf2b54/rtt/Attribute.hpp#L164) creates a shallow copy (mirror).

So I start to think that it was actually a bug.

PR #324, opened 2020-09-26
Normalize include paths

mvukov:
This should simplify compilation and make includes clear. To some extent this was necessary to build rtt with Bazel (Bazel build files are not in this PR). CMake build works on my machine as well.


mvukov:
@meyerj PTAL.

snrkiwi:
Why was the include of ```CMAKE_CURRENT_BINARY_DIR``` required? The rest of the changes look benign

mvukov:
CMake-generated config header files live in `${CMAKE_CURRENT_BINARY_DIR}/rtt` folder -- that is why I included `CMAKE_CURRENT_BINARY_DIR`.

With changes in this PR, it would be sufficient to have only two include folders: `${CMAKE_CURRENT_SOURCE_DIR}` and `${CMAKE_CURRENT_BINARY_DIR}`.

BTW, I don't think CI picked up my new commit after the forced push. The build should succeed with the latest changes.

meyerj:
Hi @mvukov! Thanks for your patch and I think it is the right approach!

Likely some more redundant `INCLUDE_DIRECTORIES()` calls can be removed in `CMakeLists.txt` files in subdirectories? For example

https://github.com/orocos-toolchain/rtt/blob/c1756f3d9b43ae08838ed0fb9613712bfa4b4bd0/rtt/CMakeLists.txt#L185-L187

https://github.com/orocos-toolchain/rtt/blob/c1756f3d9b43ae08838ed0fb9613712bfa4b4bd0/rtt/typekit/CMakeLists.txt#L12-L15

...

Would you still try and update the patch despite of the long time since you submitted this pull request? Otherwise I will try myself and update the patch soon...

> BTW, I don't think CI picked up my new commit after the forced push. The build should succeed with the latest changes.

Yes, you are right. [Both builds of this PR](https://travis-ci.org/github/orocos-toolchain/rtt/builds/730581131) have been for version fa096d092f54ba7876ac341d2d31861376123365. Anyway, the Travis configuration is currently outdated and does not work anymore since the migration to travis-ci.com. We need to update it (https://github.com/orocos-toolchain/rtt/pull/195, https://github.com/orocos-toolchain/rtt/pull/279) or switch to GitHub actions (https://github.com/orocos-toolchain/rtt/pull/327).

mvukov:
Hi @meyerj , I'll try to revive this work locally after I merge the latest master. I'll try to clean up `INCLUDE_DIRECTORIES` as well.

